PR #82, opened 2015-01-07
dblclick image opens insertImage modal; needs to return true to close modal

notno:
Otherwise, the change is saved, but the modal doesn't close.


Alex-D:
Thanks for fix. I push soon a similar solution (thanks to your, I didn't see that error). Like that : 

```js
    return $img.attr({
        src: v.url,
        alt: v.alt
    });
```

PR #88, opened 2015-02-21
Add a tbwclose Trigger

basteyy:
I think it should be part of the script.


Alex-D:
Yep great idea. Thanks!

PR #94, opened 2015-03-04
Colors: Transparent Option and Translation for Slovak

VeeeneX:
I've added Transparent option, this works like a charm but, if you change the font it turns into red color because of:
```css
font[Attributes Style] {
color: #BA0000;
}
```
And there is already Slovak translation implemented.


Alex-D:
I've added your contribution to the v2.0.0-beta.0

If you want, you can check if your patch is already OK and use the new version from bower :)

PR #97, opened 2015-03-05
Browser-based (HTML5) validation is no longer hindered by Trumbowyg

danbettles:
Hi, Alex

Thanks for sharing Trumbowyg; it's a refreshing change from the usual, bloated WYSIWYG editors.

I've made some alterations to help ensure that browser-based (HTML5) validation continues to work under Trumbowyg - textareas marked "required" continue to work as expected, at least.  

Previously, an error was being thrown by Chrome when I attempted to submit an empty Trumbowyg.  The error was occurring because the underlying textarea was marked "required" and Chrome couldn't display its validation popup because the textarea was hidden (display = none).  

The editor and textarea are no longer hidden/shown; instead, they're now stacked one on top of the other and are shown/hidden by raising/lowering z-index.  I added example files to test the alterations.

I also added some instructions to the readme on how to get started developing Trumbowyg.

Hope that helps.

Many thanks,

Dan

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,7 +7,7 @@
 /dist
 /src/ui/sass/_sprite*
 /plugins/**/ui/sass/_sprite*
-
+/nbproject/
 
 # Windows image file caches
 Thumbs.db
--- a/README.md
+++ b/README.md
@@ -18,6 +18,16 @@ http://alex-d.github.io/Trumbowyg/documentation.html
 You can contribute to Trumbowyg with translations in languages you know.
 Thanks to `node` and `gulp`, you can improve core script, style or icons easily.
 
+## Getting Started
+
+- Clone the repository.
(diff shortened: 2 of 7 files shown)


Alex-D:
Hi!

Thanks for your interest in this project, but I think it's not the slight solution for that.

One goal of this plugin is to stay below 15kb.

But I keep your PR active and I take a look ASAP to your bug.

Thanks for adding README explainations and exemples :)

Alex-D:
A big thanks to you for report this bug and propose a bugfix! I've keep your README modifications.

So, I've just released the [`v2.0.0-beta.0`](https://github.com/Alex-D/Trumbowyg/releases/tag/2.0.0-beta.0) (which weights less than 15kB :D) if you want check that the bug is correctly fixed.

Thanks again!

Issue #104, opened 2015-03-28
IME input support


foo9:
It does not work correctly on IME input. (v2.0.0-beta.1)

foo9:
References: http://www.w3.org/TR/DOM-Level-3-Events/#determine-keydown-keyup-keyCode

Alex-D:
It's seems same than #101

PR #105, opened 2015-03-28
Add IME composition

foo9:
For IME input support. #104


Alex-D:
Thanks!

I don't know how to test this fix, so I trust you :)

PR #108, opened 2015-04-05
trigger tbwchange event after syncCode

lizardK:
Hi Alex,

I moved the *tbwchange* event trigger into *syncCode* and I also call *semanticCode* into *syncCode* for 2 reasons: 

1. When you click on toolbar button, tbwchange event was not triggered
2. To avoid diff in code between the moment when event is triggered and when  semanticCode function is applied

SKÅL,

lizardK


Alex-D:
This PR is bugged, so I've patch this in a27ac98c16cc58e4aeb527654c202bba93d93ba7 and b1d63b5b8f6bc15c2a4b44a558e74143f066021a

PR #117, opened 2015-05-12
Buggy elements parsing, especially in FF (and IE 10)

Sl1v3r:
Here is what I mean: https://www.youtube.com/watch?v=GxiC1Qn3fQQ

### DIFF
--- a/src/trumbowyg.js
+++ b/src/trumbowyg.js
@@ -266,7 +266,10 @@ jQuery.trumbowyg = {
                 link: {
                     dropdown: ['createLink', 'unlink']
                 }
-            }
+            },
+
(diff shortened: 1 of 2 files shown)


src/trumbowyg.js
```diff
@@ -266,7 +266,10 @@ jQuery.trumbowyg = {
                 link: {
                     dropdown: ['createLink', 'unlink']
                 }
-            }
+            },
+
+            blockLevelElements: ['br', 'p', 'ul', 'ol', 'table', 'img', 'address', 'article', 'aside', 'audio', 'blockquote', 'canvas', 'dl', 'fieldset', 'figcaption', 'figure', 'footer', 'form', 'h1', 'h2', 'h3', 'h4', 'h5', 'h6', 'header', 'hgroup', 'hr', 'main', 'nav', 'noscript', 'output', 'pre', 'section', 'tfoot', 'video']
```
Alex-D: Why don't use string ? And why don't remove `p` from this insteed of each time `semanticCode` is called  ?

Alex-D:
Interesting PR! I will try that ASAP.

src/trumbowyg.js
Sl1v3r: I think, array in settings is more convenient and reliable, but can change this to string, if you like.
P is removed in semantics code because user can proved own settings with p in it. In that case it will break parsing, because semanticCode treats p as a special element. Moreover, I think, o.blockLevelElements may be used somewhere else, so it is expected to have fullest list of possible block-level elements, including p.

Sl1v3r:
There are still bugs, ill try another approach

PR #165, opened 2015-08-13
Add tabindex option

MiketoString:


src/trumbowyg.js
```diff
@@ -338,7 +338,8 @@ jQuery.trumbowyg = {
                 .addClass(prefix + 'editor')
                 .attr({
                     'contenteditable': true,
-                    'dir': t.lang._dir || t.o.dir
+                    'dir': t.lang._dir || t.o.dir,
+                    'tabindex': t.tabindex || -1
```
Alex-D: I'm not sure about that. By default, we don't want this attribute. Add it only if exists.

I maybe fix that ASAP or if you want, you can do it :)

MiketoString:
Hi Alex, just pushed a change to only add the tabindex attribute if the option is specified.

Alex-D:
Yeah! Can you squash your commits to keep a clean log?

Thanks :)

MiketoString:
squashed

Alex-D:
Thanks! :)

PR #179, opened 2015-09-20
Not ignore 'dist' folder

JiLiZART:
In `package.json` the `main` section pointing into `dist` folder.
When this amazing editor managed through browserify or webpack, there is a problem using it as npm module

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,7 +4,6 @@
 /uploaded-files
 /node_modules
 /bower_components
-/dist
 /src/ui/sass/_sprite*
 /plugins/**/ui/sass/_sprite*
 


Alex-D:
You need to use tagged version, not the master which is not stable. As you can see, each tagged version has the dist folder :)

https://github.com/Alex-D/Trumbowyg/tree/2.0.0-beta.5

JiLiZART:
@Alex-D npm install tagged version, but npm use .gitignore file if .npmignore not present, so, when i install through npm npm `dist` folder not present in package

Alex-D:
Oh... OK. So you could made another PR by adding a `.npmignore` that's make more sens :)

JiLiZART:
Ok, i'll try

Issue #185, opened 2015-10-06
Sprite-black/white scss not found


ra100:
I've got problem building trumbowyg from source (gulp build), sass-dist can't find imports
```@import "sprite-black"; @import "sprite-black-2x"; @import "sprite-white"; @import "sprite-white-2x";```

Am I missing something?

**Comment by Alex-D** [comment]
Need to run `gulp sprite` once

I need to fix that...

**Comment by ra100** [comment]
thanks, now it works, but I've had to first:
```npm vinyl-paths``` and ```npm gulp.spritesmith```, I know that it should be installed during ```npm install``` phase, but somehow it didn't, and running it manually worked.
Now I see that ```sprites``` task is in ```build``` task, but it takes time to execute and ```styles``` start sooner than ```sprites``` is done.

```
[08:18:13] Starting 'test-scripts'...
[08:18:14] Starting 'test-langs'...
[08:18:14] Starting 'test-plugins'...
[08:18:14] Starting 'sprites'...
[08:18:14] Starting 'styles'...
[08:18:18] Starting 'sass-dist'...
[08:18:19] Finished 'test-scripts' after 5.52 s
[08:18:19] Starting 'scripts'...
[08:18:20] trumbowyg.css all files 28.55 kB
[08:18:21] trumbowyg.js all files 39.85 kB
[08:18:21] trumbowyg.min.css all files 23.76 kB
[08:18:21] Finished 'styles' after 6.87 s
[08:18:21] Finished 'test-plugins' after 6.92 s
[08:18:21] Starting 'plugins'...
[08:18:21] trumbowyg.min.js all files 16.18 kB
[08:18:21] Finished 'scripts' after 2.48 s
[08:18:21] Finished 'sass-dist' after 2.84 s
[08:18:21] Finished 'plugins' after 160 ms
[08:18:22] Finished 'sprites' after 7.57 s
[08:18:22] Finished 'test-langs' after 7.76 s
[08:18:22] Starting 'langs'...
[08:18:22] Finished 'langs' after 168 ms
[08:18:22] Starting 'build'...
[08:18:22] Finished 'build' after 15 μs
```

ra100:
I created pull request https://github.com/Alex-D/Trumbowyg/pull/186, it should solve the problem, but I'm not sure if it's best way how to fix it

Alex-D:
Fix'd by #186 thanks to @ra100 :)

PR #186, opened 2015-10-07
fix gulp build to generate sprites before styles task

ra100:
Chained gulp task to run ```sprites``` task first, then ```styles``` and at last ```sass-dist```.
Based on http://schickling.me/synchronous-tasks-gulp/

fixes issue https://github.com/Alex-D/Trumbowyg/issues/185


Alex-D:
Thanks!

Issue #113, opened 2015-04-24
Edit link action/button


ra100:
It would be nice to have a Link edit option, just click somewhere on text which is inside <a></a> and choose from menu Edit link. And also option option to delete link without need of selecting whole text.
Maybe it could be done with plugin?

ra100:
One other think, how does "unlink" work? I'd like to add "autoextend" selection, so I can delete whole link, when I've got just cursor on link, withou need to selecting whole "linked" text.

ra100:
One other think, how does "unlink" work? I'd like to add "autoextend" selection, so I can delete whole link, when I've got just cursor on link, withou need to selecting whole "linked" text.

Alex-D:
This feature is merged now as plugin thanks to @ra100 

For auto-extend selection, It's not easy so I close this issue.

PR #187, opened 2015-10-07
editlink plugin - adds Edit Link button

ra100:
Solution to Enhancement issue https://github.com/Alex-D/Trumbowyg/issues/113

Adds option to edit link, just click on link and choose Edit Link, plugin autoselects whole link element. If you select only part of link, you will edit only part of the link. Also works for creating new links, if selection isn't link, Edit Link will create new.

next: evaluate how what should happen when selecting formatted text


Alex-D:
I need to test this ASAP and if it works I merge it :)

ra100:
with commit https://github.com/ra100/Trumbowyg/commit/b5237933146244cab65241e89ca4bf1b468acf71 I just added plugin to index.html and tidied code a bit.

Alex-D:
Thanks! It is good :)

PR #189, opened 2015-10-08
Add oEmbed plugin

jakejohns:
add plugin for http://noembed.com or compatible oembed API.
Should also work with http://iframely.com or any similar api which accepts `url` as a param and returns a response with an `html` property

### DIFF
--- a/examples/plugins/noembed.html
+++ b/examples/plugins/noembed.html
@@ -3,7 +3,7 @@
     <head>
         <meta charset="UTF-8">
         <title>oEmbed plugin | Trumbowyg by Alex-D</title>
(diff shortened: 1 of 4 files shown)


jakejohns:
In practice, I'm actually less than happy with this. The amount of markup it inserts int he editor is cumbersome. I think im finding that its easier to simply insert a link with some attribute, and deal with the actual embed on display: eg. http://sprunge.us/dgBh?javascript and https://github.com/nfl/jquery-oembed-all

Alex-D:
Your code is it stable? Can I test it? Or do you want modify it before merge? I don't understand :/

jakejohns:
The code is stable. Feel free to test it and tell me what you think.

PR #200, opened 2015-11-06
Theres no dist folder which causes npm not to work

OscarGodson:
Fix path of main JS file so that trumbowyg actually loads in Node

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "title": "Trumbowyg",
   "description": "A lightweight WYSIWYG editor",
   "version": "2.0.0-beta.5",
-  "main": "dist/trumbowyg.js",
+  "main": "src/trumbowyg.js",
   "homepage": "http://alex-d.github.io/Trumbowyg",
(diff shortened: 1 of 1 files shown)


OscarGodson:
I also realized this requires scss and it requires gulp to be run to generate sprites. Is there anyway to just get CSS, or if not, at least have the sprites built. I'm using npm so I'm not sure how I'd ever be able to deploy this since I wont be able to access `node_modules`, install gulp, run gulp sprites, and _after_ that then run npm install, start, etc on the main app.

Alex-D:
Do you get tagged version? Or trying to get master?

npm need to use this: https://github.com/Alex-D/Trumbowyg/tree/2.0.0-beta.5 for version `2.0.0-beta.5`

And you can see the dist folder.

Alex-D:
OK. I need to add a `.npmignore` file, because it's `.gitignore` which is used insteed and dist folder is ignored in it. :/

PR #207, opened 2015-12-11
Switch any remaining http to https

EHLOVader:
The gh pages page wasn't functioning through https because of the use of jquery from an insecure source.


Alex-D:
You right! Thanks! :)

PR #212, opened 2015-12-23
Upload Plugin: Patch to allow xhrFields to be added on .ajax() requests

mgoodfellow:
Hi,

This PR is to allow xhrFields to be set on the upload plugin, allowing uploads to servers with CORS and allowing authentication session information to be sent.

Example:

```javascript
$.extend(true, $.trumbowyg.upload, {
  xhrFields: {
    withCredentials: true
  },
  serverPath: 'https://my.server-on-another-domain.com/upload'
});
```


Alex-D:
Thanks! :)

PR #217, opened 2016-01-27
Allow adding classes to buttons

zerotao:
Adds CSS class support to buttons and also allows empty strings in both the button text and title.

This makes it possible to, among other things, simply override the default icons and provide icons for new buttons w/o crazy CSS/javascript hackery.

e.g.

```javascript
(function ($) {
  'use strict';

  $.extend(true, $.trumbowyg, {
    opts: {
      btnsDef: {
        table: {
          isSupported: function () {
            return true;
          },
          class: 'fa fa-table',
          title: 'Add/Edit Table',
          text: '',
          func: function (params, tbw) {
          ...
```

I also made the title setting match the new text setting for cleanliness sake.

### DIFF
--- a/src/trumbowyg.js
+++ b/src/trumbowyg.js
@@ -541,12 +541,15 @@ jQuery.trumbowyg = {
                 btn = t.o.btnsDef[n],
                 d = btn.dropdown,
                 textDef = t.lang[n] || n,
+                classDef = ((btn.class !== undefined) ? ' ' + btn.class : '' ),
+                btnText = ((btn.text !== undefined) ? btn.txt : ((btn.title !== undefined) ? btn.title : textDef)),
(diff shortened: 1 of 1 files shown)


Swift-Jr:
this is pretty much what I'm looking for! Is this going to get merged to master?

Alex-D:
I need to take a look at that this evening and check if it not break anything.

Alex-D:
I've just added this feature in this commit https://github.com/Alex-D/Trumbowyg/commit/08520187ca80872829360a991230f1d4c6272ed1

zerotao:
Ok.  I appreciate where you are going with it.  The only issue, as I see it, is that now the text segment must be an empty string which is not a huge deal but seems a little dirty compared to the empty string.

Alex-D:
Yep :( I've fail. I update that ASAP. Thanks for you review :)

zerotao:
Cool, thanks!  I'm happy to help :)

PR #219, opened 2016-02-02
add zh_cn translations to plugins

kingdido999:
I added zh_cn translations to the following plugins:

- base64
- colors
- editlink
- preformatted
- upload


Alex-D:
Thanks! :)

Issue #12, opened 2014-04-18
Button panel doesn't show the current status


ldani86:
I am not sure it is a bug report or a feature request, but it would be nice if I could see the selected text's current status on the button panel (e.g.: if I bold the selected text and then I select it again the bold button is still off)

ra100:
I have an idea how to make it work, but I don't know how to make it as plugin. I'd need to add listener on mousedown or keydown. 
Or would something like this work?
```
(function ($) {
    'use strict';

    $.extend(true, $.trumbowyg, {
          t.$ed.on('mousedown', function() {...})
    });
})(jQuery);
````

When I solve this, I can make module that will add .active class to buttons in trumbowyg panel. And also add some small css.

Alex-D:
A contributor has send me a lot of code to resolve this problem. I just need to take the time to do this.

hammat:
I'd be glad if I can be of any help !

ra100:
I've done it with plugin https://github.com/ra100/Trumbowyg/commit/3b09bdf6d2a92b09a2bc72d992bd817ebbe56c83 but I also needed to add few lines of code to trumbowyg.js https://github.com/ra100/Trumbowyg/commit/8b2b04624c45991d27e41a26bcbadcf98f5a3c45 maybe there is better solution.

Alex-D:
Note: a behavior of this issue is this sort of bug #226

ra100:
I believe 0efbbcfe184d1adc78227cec138aebf0e67f717e adds this feature to core.

Alex-D:
Yep, I've added and adapted your code to the core and make it extensible for plugins :)

PR #229, opened 2016-02-10
Add activecontrols plugin - highlights active toolbar buttons

ra100:
Plugin adds class "active" to buttons in toolbar pane based on position of cursor in editor.
Partly solves issue #12


plugins/activecontrols/trumbowyg.activecontrols.js
```diff
@@ -0,0 +1,85 @@
...
+                        if (t.selection !== null) {
+                            t.o.controlButtonActivate(t.selection.commonAncestorContainer.parentNode, t);
+                        }
+                    }
+                }
+            },
+            activeTags: [],
+            // Active tags
+            controlButtonActivate: function (element, t) {
+                var tagToButton = {
```
Alex-D: maybe add that to $.trumbowyg object to made other plugins able to add others tags for other buttons

plugins/activecontrols/trumbowyg.activecontrols.js
```diff
@@ -0,0 +1,85 @@
...
+                    }
+                }
+                t.o.activeTags = tags;
+            },
+            getTagsRecursive: function (element, tags) {
+                var tag = element.tagName;
+                if (tag === 'DIV')
+                    return;
+                if (tag === 'P') {
+                    if (element.style.textAlign !== "") {
```
Alex-D: if (tag === 'P' && element.style.textAlign !== '') {

insteed of two imbricated if would be better I think :)

Alex-D:
I test it this asap because it's a must have for a WYSIWYG. Maybe I could increase the Trumbowyg core size to add this to the core...

Thanks again!

plugins/activecontrols/trumbowyg.activecontrols.js
ra100: Right, I must have tried something where I needed nested conditions, well, seems like obsolete now.

Alex-D: No problem, I understand :)

ra100:
I'm happy that I can contribute.
And now I realized, that in case there are formatting buttons in dropdown, they are not highlighted. So maybe I should change format of `tagToButton` to use arrays, like this:
```
tagToButton = {
  'strong': ['strong'],
  'b': ['strong'],
  'h1': ['h1', 'formatting'],
```
if it's not overkill

plugins/activecontrols/trumbowyg.activecontrols.js
ra100: I'm not sure how should I do it right: `$.trumbowyg.tagToButton =` or `t.o.tagToButton`, or just add tagToButton to trumbowyg.js?

Alex-D: `$.trumbowyg.tagToButton` is better.
Look at upload plugin :)

agwidarsito:
Looks great! Let's get this in! :+1:

jakejarrett:
This would be a great addition! [LGTM](https://media.giphy.com/media/13HBDT4QSTpveU/giphy.gif)

ra100:
I see this is already implemented in core, so if I'm right, whole PR can be closed/removed.

Alex-D:
Yes! I've adapted your into the core and make it extensible for plugins :)

Thanks for your PR which help me a lot!

Issue #233, opened 2016-02-17
Insert link doesn't work on Firefox


ra100:
Tested on Firefox 44.0.2. After selecting text a clicking insert link modal doesn't show up. Just text changes to link with `href="createLink"`.
I'm looking into it, but wanted to notify about this bug.

ra100:
Seems Firefox doesn't provide in selection `extentNode`

ra100:
Fixed by 9ebe4cf4b7392b1b822fe3edace99d9d77a9b54c

Alex-D:
Fix'd by your PR :)

PR #234, opened 2016-02-17
fix insertLink and unlink in Firefox issue #233

ra100:


Alex-D:
I trust you and merge it :)

ra100:
Thanks :), I just changed selection according to https://developer.mozilla.org/en-US/docs/Web/API/Selection

PR #236, opened 2016-02-18
Added meteor support

malkesh:
Hi Alex,

Thanks for splitting branches.

Regarding builded version, I forked your repository, done required changes and created meteor package from it.

Please review changes and let me know.

Thanks,
Malkesh


Alex-D:
I keep this PR open but I think I rewrote some things before add meteor support to Trumbowyg so I prefer wait a little bit.

PR #241, opened 2016-03-03
Clarifying intent of the html getter/setter to allow setting the content to empty strings

mattsgarrison:
I think the probable intent of the `html()` method was to evaluate if the passed argument "html" was undefined and perform the getter functionality, or not undefined and perform the setter functionality. I've updated the code to explicitly check for undefined as the argument's type so that when passing an empty string the editor's content will be emptied correctly.


Alex-D:
I've just added this to develop branch because there is conflicts here and I prefer `!= null` test.

Thanks for you PR!

PR #249, opened 2016-04-07
Remove id property from insertHorizontalRule

akupiec:
Currently insertHorizontalRule are adding also id="null" property to DOM with is pointless.

This small change fix it.


Alex-D:
Thanks for this patch :)

Issue #256, opened 2016-04-15
Button icons are now shown with base href in html


Aleksandr-ru:
Hello :)
If there is a ````<base href="...">```` element in HTML document, pointing with href to somewhere else than current url, there are no icons visible on buttons of the editor.

This is caused by ````<svg><use xlink:href="...">```` in buildBtn function.

My way to fix it is to add ````window.location```` to the href, for example add it to line 675:
`````
html: '<svg><use xlink:href="' + window.location + '#' + prefix + (btn.ico || btnName).replace(/([A-Z]+)/g, '-$1').toLowerCase() + '"/></svg>',
````
And line 738:
````
html: '<svg><use xlink:href="' + window.location + '#' + prefix + (btn.ico || btnName).replace(/([A-Z]+)/g, '-$1').toLowerCase() + '"/></svg>' + (btn.text || btn.title || t.lang[btnName] || btnName),
````
Or maybe ````window.location.href.split('#').shift()```` as a 100% solution
Thanks)

### Informations

**Browser + version**: Firefox 45 or any
**OS**: Windows 7 or any
**Resolution**: 1680x1050 or any


### How to reproduce the bug?
Make a page in a different path than Trumbowyg
Add Trumbowyg
Add  ````<base href="...">```` pointing to somewhere different than the page
See that there are no icons

RespectTheCoder:
Can confirm that this solution works.

SwithFr:
I can confirm too ;)

Miztah87:
Thanks @Aleksandr-ru  , great solution.
 Works like a charm!

lucidlemon:
uhm could we get this integrated by default? makes it a bit hard to work with browserify right now, as I don't want to manipulate files within the node_modules folder.

Alex-D:
I need to update the doc for the v2. In fact, you can use the `svgPath` option, but that does not fix this bug. If you apply the OP bugfix you break the IE/Edge support.

Alex-D:
Pretty patch: https://gist.github.com/leonderijke/c5cf7c5b2e424c0061d2

Alex-D:
I've just work on a patch on the `develop` branch. If someone want to test, he is welcome :)

Alex-D:
Fix'd in v2.1.0. See https://alex-d.github.io/Trumbowyg/documentation.html#svg-icons if you need it

PR #258, opened 2016-04-21
Fix issue #256

SwithFr:
fix issue [256](https://github.com/Alex-D/Trumbowyg/issues/256)


Alex-D:
I don't want do that because it breaks support of IE 9-11. See 3 point of Know issues pane here: http://caniuse.com/#feat=svg

Thanks for your purpose but that's not the solution for me.

SwithFr:
Oh yes I understand ;)

PR #260, opened 2016-04-21
Allow custom styling of Buttons via CSS

luetge:
In the past, I used to replace the default icons by FontAwesome icons in order to have matching icons on the whole page. In version 2, this is not possible anymore because one is forced to use the SVG icons. This simple change allows to disable the default usage of SVG Images for the icons by setting the option `svgPath` to `false`. Furthermore, the `$.get(svgPath, ...)` should be made safer to handle badly formed data (XMLSerializer complains in FF if data.documentElement is null or undefined).

### DIFF
--- a/src/trumbowyg.js
+++ b/src/trumbowyg.js
@@ -140,7 +140,7 @@ jQuery.trumbowyg = {
         }
 
         // SVG path
-        if ($('#' + trumbowygIconsId, t.doc).length === 0) {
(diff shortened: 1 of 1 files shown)


Alex-D:
That's a good idea! Thanks :)

luetge:
Thanks for merging and providing such a slick widget, very helpful! :thumbsup:

PR #261, opened 2016-04-23
added ability to insert audio elements

AdamHess:
This allows you to insert audio elements into the editor, this will be extended to allow base64 encoded audio in the future. 

Please recommend an alternative icon as the one I am proposing is sourced from: 

https://github.com/frexy/glyph-iconset/blob/master/svg/si-glyph-document-music.svg

Which may require changing the licensing type if used.


AdamHess:
found a better licensed Icon (MIT) will reopen using different pull request.

PR #262, opened 2016-04-23
Added ability to insert audio sources

AdamHess:
This allows you to insert audio elements into the editor, this will be extended to allow base64 encoded audio in the future.

Icon is sourced from https://useiconic.com/open/ (MIT licensed)


Alex-D:
Thanks for your PR. But I don't think so this feature had his place in the core.

If you could transform this code into a plugin, it could be merge :)

AdamHess:
Unfortunately, I don't have much more time to volunteer to the project.
I'll just keep my fork.

--A

Alex-D:
Let that PR open, I will take a look at that to make a plugin if I took the time :)

PR #282, opened 2016-05-27
Insert Audio

AdamHess:
This included adding "checkbox" options (and resizing of labels so that they display correctly)

Complete with an example page 

This is to be a one of many plugins for inserting media.


Alex-D:
I've released your plugin in [v2.2.0](https://github.com/Alex-D/Trumbowyg/tree/v2.2.0)
Thanks for your contribution!

PR #297, opened 2016-06-09
Fixes exception if script is loaded with search query

TikiTDO:
If the script tag used to load the script had a search query, then this code would cause an exception. This resolves the problem


Alex-D:
Ok, I see the problem. Thanks :)

PR #301, opened 2016-06-14
Added the ability to have custom icons for custom buttons.

NoczCore:
**Hi Alex,**

I added the ability to have custom icons on custom buttons. This may be icons as an image or a class of operating with libraries as [Ionicons](ionicons.com) or [Font Awesome](fontawesome.io).

## To put an image as an icon:
```
$.extend(true, $.trumbowyg, {
            plugins: {
                somePlugin: {
                    init: function(trumbowyg){
                        var btnDef = {
                            iconUrl: 'http://domain.ext/path/../myIcon.png', // PNG, JPEG, SVG, etc..
                            fn: function () {
                                ...
                            }
                        };
                        trumbowyg.addBtnDef('somePlugin', btnDef);
                    }
                }
            }
});
```

**Will generate:**
```
<img src="http://domain.ext/path/../myIcon.png">
```

## To use a library such [Ionicons ](ionicons.com)or [Font Awesome](fontawesome.io):
```
$.extend(true, $.trumbowyg, {
            plugins: {
                somePlugin: {
                    init: function(trumbowyg){
                        var btnDef = {
                            iconClass: 'ion ion-ionic', // Ionicons, Font Awesome and others...
                            fn: function () {
                                ...
                            }
                        };
                        trumbowyg.addBtnDef('somePlugin', btnDef);
                    }
                }
            }
});
```

**Will generate:**
```
<i class="ion ion-ionic"></i>
```

Thank you for your work,
_Sincerely, NoczCore._

### DIFF
--- a/src/trumbowyg.js
+++ b/src/trumbowyg.js
@@ -668,27 +668,36 @@ jQuery.trumbowyg = {
                 btn = t.btnsDef[btnName],
                 isDropdown = btn.dropdown,
                 textDef = t.lang[btnName] || btnName,
+                html = '';
 
-                $btn = $('<button/>', {
-                    type: 'button',
-                    class: prefix + btnName + '-button ' + (btn.class || ''),
-                    html: t.hasSvg ? '<svg><use xlink:href="' + t.svgPath + '#' + prefix + (btn.ico || btnName).replace(/([A-Z]+)/g, '-$1').toLowerCase() + '"/></svg>' : '',
-                    title: (btn.title || btn.text || textDef) + ((btn.key) ? ' (Ctrl + ' + btn.key + ')' : ''),
-                    tabindex: -1,
-                    mousedown: function () {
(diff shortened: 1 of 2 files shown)


Alex-D:
Hi!

There is already a `class` option to add a custom class to a button. But it seems like this option is missing on sub buttons (in dropdown) and is not documented I think.

If you want add png icon or whatever you should use CSS, not JS/HTML. I need to insert SVG icons like that to benefits of SVG sprite, there is no other solution at the moment that works on IE/Edge and maybe other browsers.

Your code is sooooo heavy/huge, and add a lot of line to a plugin that want to be light :/

So, I check asap to add the class option to sub buttons that is one of your features, and I let you go in your CSS to add some custom icons. Set [svgPath option to false](http://alex-d.github.io/Trumbowyg/documentation.html#svg-icons) if you don't want SVG icons.

Thanks for your suggestion anyway :)

PR #306, opened 2016-06-18
Events added

hiendv:
`onSuccess` & `onError` added

### DIFF
--- a/plugins/upload/trumbowyg.upload.js
+++ b/plugins/upload/trumbowyg.upload.js
@@ -20,7 +20,9 @@
         urlPropertyName: 'file',
         statusPropertyName: 'success',
         success: undefined,
-        error: undefined
+        error: undefined,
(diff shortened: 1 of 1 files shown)


Alex-D:
Hi,
this is not an event, that is a callback.

There is already a `success` and `error` functions. I'm OK if you want trigger an event, but not for callback, sorry.

hiendv:
@Alex-D Hi, by saying "Events" I meant the callbacks for the corresponding events.
The problem with the current implementation is: I don't want to override the default behavior of these callbacks, I just want to have "my callbacks" dispatched when the default methods are finished.

hiendv:
So I should make another PR which triggers these events or it's already implemented?

Alex-D:
You can make a PR, that would be awesome :)

hiendv:
@Alex-D Please check my latest PR #309 ;)

PR #307, opened 2016-06-20
Only disable image resize in Firefox (enableObjectResizing). Fixes #278.

### LINKED ISSUE #278: enableObjectResizing not supported in IE10- browsers
Trying to implement in our intranet - which unfortunately requires IE10-. It's inside a bootstrap modal box and when initializes, it breaks at line 447:

 **t.doc.execCommand('enableObjectResizing', false, false);**

trumbog1

It's a known bug: https://dev.ckeditor.com/ticket/4208, but I'm wondering how I can fix so it is ignored in IE10. 
### Informations

**Browser + version**: IE 10-
**OS**: Windows 7
**Resolution**:

### How to reproduce the bug?

I have a jsfiddle: https://jsfiddle.net/mks6804/rbk7sgkr/7/ - again, it has to run in stupid IE10-.  

Any help is appreciated.

btecu:
Only Firefox supports the `enableObjectResizing` command. Fixes crash in IE10.

### DIFF
--- a/src/trumbowyg.js
+++ b/src/trumbowyg.js
@@ -66,6 +66,9 @@ jQuery.trumbowyg = {
 (function (navigator, window, document, $) {
     'use strict';
 
+    // Browsers
+    var isFirefox = typeof InstallTrigger !== 'undefined';
+
     $.fn.trumbowyg = function (options, params) {
         var trumbowygDataName = 'trumbowyg';
         if (options === Object(options) || !options) {
@@ -433,7 +436,10 @@ jQuery.trumbowyg = {
             t.initPlugins();
 
             // Disable image resize in Firefox
-            t.doc.execCommand('enableObjectResizing', false, false);
+            if (isFirefox) {
+                t.doc.execCommand('enableObjectResizing', false, false);
+            }
+
             t.doc.execCommand('defaultParagraphSeparator', false, 'p');
 
             t.buildEditor();


Alex-D:
Why not just try/catch?

btecu:
Because that's bad programming:

- try catch adds overhead (execute statement, generate error, log error, make stacktrace, move back to code)
- exceptions should be used for exceptional circumstances

In this case, it doesn't make sense to run something on all browsers when only Firefox supports it - even if it doesn't throw on all browsers.

Alex-D:
Since `enableObjectResizing` is an official execCommand option and IE/Edge seems support it, I go to try catch it because all browsers will support it one day.

More: https://connect.microsoft.com/IE/Feedback/Details/791253

btecu:
I guess you haven't bothered to read. The link says nowhere that it was implemented. Moreover, MSFT offers an alternate solution:
```
function controlselectHandler(evt) {
  evt.preventDefault();
}

document.body.addEventListener('mscontrolselect', controlselectHandler);
```

Anyway, terrible approach with try/catch.

Alex-D:
> terrible approach with try/catch.

This command is run once at the creation of the editor and only throw error on IE10- and should be implemented in all browsers in a perfect world. I don't want to need to maintain a list of browser that use that or not in each case.

I totally understand you, but in this case the problem is IE, not my code. I should not have to catch this edge case...

Anyway, thanks for your PR, it was appreciable :)

btecu:
My point was that `enableObjectResizing` is only available in Firefox. It's more likely to be deprecated from Firefox than to be implemented by another browser. So no, the problem is not really IE.

PR #308, opened 2016-06-20
Handle Enter for IE and Edge. Fixes #305.

### LINKED ISSUE #305: Entering new line only works the first time on IE
Using IE11 and Edge, if you press enter once, it works correctly adding a new line.
However, if you press enter again, it does not add a new line. Looking at the source, in IE it's adding `&nbsp;` for enter instead of `<br>` (in Chrome).

btecu:
Handle IE and Edge adding `&nbsp;` instead of `<br>`.

### DIFF
--- a/src/trumbowyg.js
+++ b/src/trumbowyg.js
@@ -66,6 +66,10 @@ jQuery.trumbowyg = {
 (function (navigator, window, document, $) {
     'use strict';
 
+    // Browsers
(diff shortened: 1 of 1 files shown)


Alex-D:
Wow this is huge just for IE :(
I think we can forget IE10, only IE11 and Edge are to support for me.

btecu:
Dropping `IE10` would save about four lines of code in this case.
While size is important, correct functionality is more important.

PR #309, opened 2016-06-21
Upload events added

hiendv:
An alternative to #306

### DIFF
--- a/plugins/upload/trumbowyg.upload.js
+++ b/plugins/upload/trumbowyg.upload.js
@@ -146,13 +146,13 @@
                                                     setTimeout(function () {
                                                         trumbowyg.closeModal();
                                                     }, 250);
-                                                    trumbowyg.$c.trigger('upload.success', [trumbowyg, data, url]);
+                                                    trumbowyg.$c.trigger('tbwuploadsuccess', [trumbowyg, data, url]);
                                                 } else {
                                                     trumbowyg.addErrorOnModalField(
                                                         $('input[type=file]', $modal),
                                                         trumbowyg.lang[data.message]
                                                     );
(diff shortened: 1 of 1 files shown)


hiendv:
As you can see `data` is optional and could be omitted on `upload.error`
An example (ES6)
```js
$(node).trumbowyg()
.on('upload.success', (trumbowyg, data, url) => {
    console.log(trumbowyg, data, url)
})
.on('upload.error', (trumbowyg, data) => {
    console.log(trumbowyg, data)
})
```

Alex-D:
Can you follow the core event name "normalization"?
See here: http://alex-d.github.io/Trumbowyg/documentation.html#events

Something like `tbwuploadsuccess` and `tbwuploaderror` maybe

Alex-D:
Thanks! And sorry to be exigent ^^'

hiendv:
@Alex-D I changed the names and it's my fault for not following the convention. It's okay. Anyway, what are your opinions on another naming convention for events? Let's say `namespace.plugins.event`. E.g. `tbw.upload.success`

Alex-D:
I've tested them, but jQuery has his own event namespace management which is strange and not vanilla compatible (or I was wrong).

hiendv:
@Alex-D I didn't know the difference. jQuery even supports [`event.namespace`](https://api.jquery.com/event.namespace/) or am I missing something?

Alex-D:
jQuery yep, but it's not a JavaScript standard. But I need to take a look over that again.

hiendv:
@Alex-D Totally understand. We will talk about it later, in another discussion ;).
Targeting to 2.2.x or 3.x.x maybe :smile:

PR #312, opened 2016-06-23
Fixing svgPath when we are using anchors

rwrz:
When using anchor on the application, like http://example.com/example#test, svgPath returns everything, including anchors, resulting in a wrong svg load. With this fix, we don't get anchors anymore, fixing the svg import.


Alex-D:
Thanks!

Issue #353, opened 2016-08-02
Strikethrough cannot be undone


sbunciak:
Before report an issue, be sure to put these informations. Thanks :)

### Informations

**Browser + version**: Safari 9.1.2, Chrome
**OS**: Mac OS X El Capitan
**Resolution**:

### How to reproduce the bug?
Select text in editor (even demo on your page), make it strikethrough. There is no way of un-doing it. Clicking again on the 'strikethrough' nor 'remove formatting' button work.

L3CC:
As a temporary solution, it will work if you set 'semantic: false'.

PR #356, opened 2016-08-04
Workaround document.execCommand doesn't recognize <del> tag. Fixes #293 #353.

### LINKED ISSUE #293: Can't "turn off" <del>
It seems that it's the only tag where this happens (strong, italic, etc, work fine).

rsusanto:
Firefox, Chrome and Safari are consistently using `<strike>` when formatting strikethrough with execCommand. I guess it is safe to use this workaround.

### DIFF
--- a/src/trumbowyg.js
+++ b/src/trumbowyg.js
@@ -1193,6 +1193,13 @@ jQuery.trumbowyg = {
                         param = '<' + param + '>';
                     }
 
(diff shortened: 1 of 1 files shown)


Alex-D:
This is a non-sense with that: https://github.com/Alex-D/Trumbowyg/blob/develop/src/trumbowyg.js#L1004

Issue #313, opened 2016-06-24
CTRL + X  is not reflecting on text-area(#game_description). 


rahul8sharma:
```js
    $('#game_description').trumbowyg({
        resetCss: true,
        removeformatPasted: true,
        autogrow: true,
        svgPath: trumbowygIcons
    }).on('tbwchange', function(contents, $editable){
        $('#game_description').val(contents.target.value);
    });
```

Whenever i pressed CTRL + X then Textarea not updated.

rahul8sharma:
Still not fixed because it cut value from the editor after "tbwchange" event performed.  So "contents.target.value" return the same value as previous.

Alex-D:
Need to add a setTimeout 0 hack :/

rahul8sharma:
We can use
.bind('input propertychange', function() {
    t.semanticCode(false, true);
    t.$c.trigger('tbwchange');
})

instead of 

.on('cut', function () {
    t.semanticCode(false, true);
    t.$c.trigger('tbwchange');
})

manaswigirlscollege:
Thank you.. :) Now it is working.

supercharlez:
Hello, I don't understand how to fix this issue I have too with Ctrl-X not updating the content. Can someone help ? thanks

Alex-D:
It should works in the latest version. Are you up to date?

supercharlez:
Yes silly me I was using the last version on a cdn and not the one on Github. Thank you so much

PR #358, opened 2016-08-10
It will solve CTRL + X issue

rahul8sharma:
https://github.com/Alex-D/Trumbowyg/issues/313


Alex-D:
Can't works.
- prefer `on` over `bind`: http://api.jquery.com/bind/#entry-longdesc
- `input` event is fired a lot, not only on cut

rahul8sharma:
But "cut" event trigger before cut the text. So we can not replace cut by bind event but we can use instead of so many events.

PR #368, opened 2016-08-25
Add accepts-attribute to image upload.

Droeftoeter:
This will give compatible browsers a hint to only allow images.


Alex-D:
Thanks :)

PR #371, opened 2016-08-25
I corrected some errors in translation

chadidi:


Alex-D:
You should fix that from and to `develop` branch.

PR #374, opened 2016-08-30
Mime-type validation for base64 inline images, template plugin.

Droeftoeter:
Added mime-type validation to the base64 plugin, it will now display an error if you select a file that is not an image.

Added a very basic template-plugin, select a template and it will replace the current editor content with the content the template. Because it is very difficult to find an icon for template, I also made it possible to have textual-buttons in the button-bar.

### DIFF
--- a/plugins/template/trumbowyg.template.js
+++ b/plugins/template/trumbowyg.template.js
@@ -1,0 +1,52 @@
+(function($) {
+    'use strict';
+
+    // Adds the language variables
+    $.extend(true, $.trumbowyg, {
+        langs: {
+            en: {
+                template: 'Template'
+            },
+            nl: {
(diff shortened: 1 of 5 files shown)


plugins/template/trumbowyg.template.js
```diff
@@ -0,0 +1,52 @@
+(function($) {
```
Alex-D: Can you move this file in `exemples` folder?

Droeftoeter:
@Alex-D I've added an example for the template plugin.

Alex-D:
Thanks!
I just need to check that when I go back home, then I should merge :)

Alex-D:
Cool! It works well!

PR #375, opened 2016-08-30
A few mistakes in the translation corrected

chadidi:
I've been corrected a few mistakes in the translation of ar.js file


Alex-D:
Duplicated of #376

PR #378, opened 2016-08-31
When editing text in the trumbowyg controllers the arrow keys navigat…

rahul8sharma:
When editing text in the trumbowyg controllers the arrow keys navigate the tabs and not the text. This happened when we navigate the arrow keys it propagate the all parents. So, I have added stopPropagation() to stop parent propagation.

### DIFF
--- a/src/trumbowyg.js
+++ b/src/trumbowyg.js
@@ -557,6 +557,7 @@ jQuery.trumbowyg = {
                     }
                 })
                 .on('keyup input', function (e) {
+                    e.stopPropagation()
                     if (e.which >= 37 && e.which <= 40) {
                         return;
                     }


Alex-D:
Didn't understand the fix. 
Arrow keys are already catched by the code below: 

```
if (e.which >= 37 && e.which <= 40) {
    return;
}
```

rahul8sharma:
That arrow key use to navigate the text inside the editor but if i have tab system in my application then i will press arrow key like left or right it will navigate a tabs. so i have stopped propagation and it will not navigate parent.

Alex-D:
OK, I have reproduce the bug, your patch is not the good way to fix the problem. The problem is: the content editable field keep like a focus and need to be blur.

That's not easy :/

geektortoise:
I had the same problem and I've shamefully put the same line here :

https://github.com/Alex-D/Trumbowyg/blob/develop/src/trumbowyg.js#L657

By the way, thanks for this awesome lib ! :-)

PR #379, opened 2016-08-31
Font Family and Font Size plugins

Droeftoeter:
I've added two extra buttons, one is to select a font-family from a list of websafe fonts. and the other lets you select a font-size from a set of predefined sizes (xs, s, m, l, xl). Examples included.


src/trumbowyg.js
```diff
@@ -1486,6 +1486,37 @@ jQuery.trumbowyg = {
         getRangeText: function () {
             return this.range + '';
         },
+        expandRange: function () {
```
Alex-D: I'm not OK to add a *huge* function to core for that :/
Need to optimize the size...

src/trumbowyg.js
```diff
@@ -1486,6 +1486,37 @@ jQuery.trumbowyg = {
...
+                documentSelection = t.doc.getSelection();
+
+            if (documentSelection.rangeCount) {
+                var ranges = [];
+                for (var i=0; i<documentSelection.rangeCount; i+=1) {
+                    var range = documentSelection.getRangeAt(i);
+                    var ancestor = range.commonAncestorContainer;
+
+                    if (ancestor.nodeType === 1) {
+                        if (range.startContainer.parentNode !== ancestor && documentSelection.containsNode(range.startContainer.parentNode, true)) {
```
Alex-D: Use a variable to store `range.startContainer.parentNode` because like that, uglifier can't minify this code.

So the minified from this code should be something like this:

```
if (r.startContainer.parentNode !== ancestor && d.containsNode(r.startContainer.parentNode, true)) {
```

But if you use a variable to store the parentNode it should be minified to something like that:

```
var p = r.startContainer.parentNode;
if (p !== ancestor && d.containsNode(p, true)) {
```

And if you see other optimizations like that you can apply them.

The gulpfile log the trumbowyg.min.js size and this size should not go over 21kB

src/trumbowyg.js
Alex-D: What do `expandRange`? Extend to the first block?

saggiyogesh:
Just a thought. 
Is it not possible a package these as separate plugins just like `color` plugin ?

Alex-D:
I think that should be in each plugin, but if this is a common part, maybe it could be interesting to add that method to the core... but, maybe it could be interesting to create a plugin that add that method to the core.

Something like a `expendSelection.js` that just register this new method and which is needed by the plugins that need this function.

```js
$.extend(true, $.trumbowyg, {
    plugins: {
        _expandSelection: {
            init: function (trumbowyg) {
                trumbowyg._expandSelection = function () {
                    // The code here...
                };
            }
        }
    }
});
```

Droeftoeter:
I'll take a look at decreasing the size. The reason I put it in the core is because I can imagine more plugins in the future will want to select the full text node or full node when applying an action.

Droeftoeter:
I tried to optimize it. It's a bit smaller now.

ffaubry:
@Alex-D Are you going to merge?

Alex-D:
I don't think so, because of the new fat function added to the core.

hytromo:
While I completely understand the reason why this hasn't merged yet, if anyone still tries to hack this into their project (like what I'm trying to do), then I've found that changing the `expandRange` function to do 

```
range.setStart(ancestor, range.startOffset);
range.setEnd(ancestor, range.endOffset);
```
under `else if (nodeType === 3)` to give a much closer feel to what office applications do (only the selected text changes, not everything under the same parent).

src/trumbowyg.js
PokatilovArt: Seems, that it takes expandes selected range till beginning and ending of paragraph.
For example, if you select the word "Ipsum" in paragraph "Lorem ipsum — dolor sit amet, consectetuer adipiscing elit", then you will get all the sentence.

But I don't understand, why do we need it. I guess, that it's a specific task for a commited user, but it without it plugin should work more correct. 

That is why:
LOC 52  `trumbowyg.expandRange();` in plugins/fontsize/trumbowyg.fontsize.js should also be remove.
>

floribon:
+1 for those plugins to be merged after cleanup.

I agree that **expandRange** should not be part of this PR as it can apply to any other formatting and in my case i don't need it, I want to change the font size only of the part being selected (I just commented out the call to `trumbowyg.expandRange()` and it works).

Droeftoeter:
I've removed the **expandRange** function from the core and synced the branch.;

Alex-D:
Can someone add some documentation into docs folder?

PR #381, opened 2016-09-02
Updating semantic to allow for more refined control over which elemen…

nurogenic:
…ts get mapped. Also allows people to refrain from forcing div tags to become p tags which was the inital reason for the update.

### DIFF
--- a/src/trumbowyg.js
+++ b/src/trumbowyg.js
@@ -344,11 +344,7 @@ jQuery.trumbowyg = {
 
             prefix: 'trumbowyg-',
 
-            semantic: {
-              'b': 'strong',
-              'i': 'em',
-              'div': 'p'
-            },
+            semantic: true,
             resetCss: false,
             removeformatPasted: false,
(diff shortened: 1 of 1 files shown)


Alex-D:
That could be interesting with some code optimizations yep. I don't have the time yet, but I should take a look at that soon.

nurogenic:
Ok great. Let me know if there's something you'd like me to take care of if you don't have the time.

richardtallent:
This looks similar to a need I have, but I'm not sure -- in my application, I actually *want* B, I, U, S, SUB, SUP, CODE, and BR to be allowed (along with OL, UL, and LI), but I don't want ANY other markup to be permitted (or any attributes), and any DIV or P should be converted to a trailing BR (ideally unless at the end or followed by OL/UL).

I know my needs are unusual, but I'm hoping for a generalized solution to transforming and whitelisting tags. Will this branch get me closer, or should I create a separate issue?

PR #385, opened 2016-09-06
Update trumbowyg.js

adamdburton:
Fixes the removeformat button doing nothing by creating a temporary div, setting the html for the div, grabbing the text value and replacing newlines with `<br>`'s. Then formats semantically.

Definitely not perfect but removes all formatting.

### DIFF
--- a/src/trumbowyg.js
+++ b/src/trumbowyg.js
@@ -917,6 +917,17 @@ jQuery.trumbowyg = {
         },
 
 
+        removeformat: function () {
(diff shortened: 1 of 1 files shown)


Alex-D:
I didn't understand what doesn't works?
I've retry on the [presentation page](http://alex-d.github.io/Trumbowyg/) with latest Chrome, all it's OK.

What's your browser? Give me more informations please :/

PR #387, opened 2016-09-06
Namespacing events

nurogenic:
Ran into issues where it was removing all click events from the body so i'm namespacing all on() and off() events.


nurogenic:
Ah I must have branched from master by mistake. The only changes I made were to src/trumbowyg.js

PR #388, opened 2016-09-06
Ran into issues where it was removing all click events from the body …

nurogenic:
Ran into issues where it was removing all click events from the body so i'm namespacing all on() and off() events.

### DIFF
--- a/src/trumbowyg.js
+++ b/src/trumbowyg.js
@@ -424,6 +424,9 @@ jQuery.trumbowyg = {
             t.addBtnDef(btnName, btnDef);
         });
 
+        // put this here in the event it would be merged in with options
+        t.eventNamespace = 'trumbowyg-event';
+
         // Keyboard shortcuts are load in this array
         t.keys = [];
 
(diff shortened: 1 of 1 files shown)


Alex-D:
I did not understand the bug :/

It bug when Trumbowyg do off ? Or when you do off in your app ?

nurogenic:
Trumbowyg was removing all events from the body even the ones that didn't have to do with trumbowyg.

Alex-D:
The only one off on body is here: https://github.com/Alex-D/Trumbowyg/blob/develop/src/trumbowyg.js#L964

Can you confirm me that's the only one event that was disabled by Trumbowyg?

nurogenic:
Correct but you're also adding events to the window and never removing them which is another issue. But If I wanted to add click events in my app on top of the events you have in trumbowyg I can't trust that you will only remove your events and not the ones I've added after the fact. Its good practice to namespace all events in a plugin to allow for developers to add their own code ontop without running into conflicts. If you wanted to break it down further and have form event namespace and a global event namepsace thats up to you but you should namespace all events.

Alex-D:
Yep but namespaces are a jQuery feature (or I missed something?) and I want to stay vanilla-compatible at the maximum.

But if it is necessary, I will add some namespaces :/

nurogenic:
I would say just add it to the body and the window since they're global elements for now as a bare minimum.

nurogenic:
I updated the pull request to only add namespacing to the window and body as well as add an off() call to remove events from the window on destroy.

https://github.com/Alex-D/Trumbowyg/pull/388/files#diff-a57e0e52aed1451208c5f2010cd3b537R1521

This line is fixing another issue where the element didn't have a parent element so I'm just checking one exists before setting it. I can separate out if you'd like it was committed by mistake.

Alex-D:
Sorry I have forget this PR. Thanks :)

PR #392, opened 2016-09-07
Recommend npm before bower

dandv:
...based on [the ecosystem moving towards npm](http://stackoverflow.com/questions/18641899/what-is-the-difference-between-bower-and-npm/31219726#31219726).


Alex-D:
OK. Why not. Thanks :)

PR #408, opened 2016-09-27
Plugin to clean pasted content from Microsoft Word

eradin:


Alex-D:
Thanks! I will test that.

PR #416, opened 2016-10-07
Little bug fix when using the *tbwchange* event

netbull:


Alex-D:
You PR is dirty :(
Your fix is not a correct fix.
What's the problem that you want solve?

netbull:
When using it like this:

`$('#editor')
    .trumbowyg()
    .on('tbwchange', function(){ console.log("click"); });`

the event is triggered two times in a row

Alex-D:
Use a throttle :)

netbull:
to use what ? O_O

Alex-D:
If you want to avoid the bounce effect, use a debouce or throttle method.

netbull:
the debounce method doesn't work.. the event is still called two times

Alex-D:
Yep, but if you don't want to do 2 times an action, you just have to debounce it.

```
$('#your-editor').trumbowyg();
$('#your-editor').on('tbwchange', _.debounce(function () {
    console.log('CHANGE!');
}));
```

netbull:
Yep, this is exactly what I did :) even I test if I initialize the editor twice by mistake or something, but no.. it's initialized only once, but with, or without the debounce the event is fired twice.

The editor is initialized in a Backbone View, but now I'm experimenting to switch from Backbone to ReactJS, so will see there how it will handle.

PR #420, opened 2016-10-17
Add vuejs directive to GitHub page

Elhebert:
For one of my project, I did a vuejs 1.0 implementation of Trumbowyg ([https://github.com/Elhebert/v-trumbowyg](https://github.com/Elhebert/v-trumbowyg)).

I took the liberty to add it to gh-page for other people to be able to use it.

If you have any questions, I'm here.

### DIFF
--- a/index.html
+++ b/index.html
@@ -606,6 +606,12 @@ $('#trumbowyg-demo').trumbowyg({
                             AngularJS directive
                         </a>
                     </li>
+                    <li>
+                        <a href="https://github.com/Elhebert/v-trumbowyg">
+                            <img src="img/packages/vue.svg" alt="">
+                            VueJS 1.0 directive
+                        </a>
+                    </li>
                 </ul>
             </div>
         </div>


Alex-D:
Thanks! I will add the icon :)

Alex-D:
Your directive is not compatible with Vue 2 ? :/

Elhebert:
For the icon, i did a quick one based on a vuejs svg logo. But design isn't my strong suits 😄 So, I'm sure you'll be able to do a better one 👍 

For the compatibility, I'm not sure, it is supposed to be (according to the patch notes, nothing there has changed), but I have to test it to be sure.

I'll try to do this tonight.

Alex-D:
If you test it on Vue.js 2 tell me if it works on both versions. We can update to drop the "1.0" precision on the Trumbowyg page :)

Elhebert:
I did a quick check in the documentation, and, I was wrong, they did quite a lot of changes for the directives in VueJS 2.0, especially for the TwoWay bindings directives.

I'll need to transform it into a component, thus, it won't be compatible with VueJS 1.0.

I don't know if github support links to another branch inside a README, if they do, I'll just do antoher branch for VueJS 2.0, with a link to it in the README.

If they don't support it, I'll do 2 folders, one for each version, with separate docs, and a main README to redirect a user to the correct version.

Either way, you'll have a single link on the Trumbowyg page.

Alex-D:
Yep, you just have to take a link like that:

`https://github.com/Alex-D/Trumbowyg/tree/master`

PR #424, opened 2016-10-21
Update trumbowyg.emoji.js

NicolasPion:
Fix bug / bad syntax on insert


Alex-D:
Thanks for your fix! :)

Issue #435, opened 2016-11-04
DefaultParagraphSeparator not supported in IE < 11


dominiczaq:
### Informations

**Browser + version**: IE9 / IE10
**OS**: Windows 7
**Resolution**: Not relevant

### How to reproduce the bug?

Open webpage with trumbowyg. IE < 11 are not supporting DefaultParagraphSeparator https://msdn.microsoft.com/en-us/library/hh801229(v=vs.85).aspx#DefaultParagraphSeparator

dominiczaq:
I've made PR with simple try catch block, which solves the issue.

PR #436, opened 2016-11-04
Wrap defaultParagraphSeparator execCommand in try catch block

dominiczaq:
DefaultParagraphSeparator is not supported in IE < 10
https://msdn.microsoft.com/en-us/library/hh801229(v=vs.85).aspx#DefaultParagraphSeparator

Fixes issue https://github.com/Alex-D/Trumbowyg/issues/435

### DIFF
--- a/src/trumbowyg.js
+++ b/src/trumbowyg.js
@@ -452,7 +452,11 @@ jQuery.trumbowyg = {
                 t.doc.execCommand('enableObjectResizing', false, false);
             } catch (e) {
             }
-            t.doc.execCommand('defaultParagraphSeparator', false, 'p');
+            try {
+                // Disable for old IE (support from IE 11)
+                t.doc.execCommand('defaultParagraphSeparator', false, 'p');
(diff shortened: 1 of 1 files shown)


src/trumbowyg.js
```diff
@@ -452,7 +452,11 @@ jQuery.trumbowyg = {
                 t.doc.execCommand('enableObjectResizing', false, false);
             } catch (e) {
             }
-            t.doc.execCommand('defaultParagraphSeparator', false, 'p');
+            try {
+                // Disable for old IE (support from IE 11)
+                t.doc.execCommand('defaultParagraphSeparator', false, 'p');
```
Alex-D: This should be in the same try catch as the execCommand above, if the first fail, the second one will fail too.

dominiczaq:
I didn't initially look up support for enableObjectResizing. Just as you said, both lines should be in same try catch block.

Alex-D:
Thanks! :)

Issue #411, opened 2016-09-29
What is the composition variable for? (trumbowyg.js:542)


hurricane766:
There's a demo here: https://api.jquery.com/event.which/ , that will log e.which as you type, and when using trumbowyg in Chrome on an Android device, the 229 code is fired for each key press (except enter). 

I tracked the issue back to trumbowyg.js:568 and the composition boolean variable. Because the 229 keydown is detected, composition is set to true on line 548 and then line 568 prevents regular key presses from firing the tbwchange event.

Alex-D:
This event is for languages that need multiple keys to write one symbol like chinese or some languages with non-latin symbols.

I know it's buggy on Android, maybe I need to make an option to disable that...

hurricane766:
Perhaps an option you could add to the language definitions? So if the language needs multiple keys it could be turned on.

Alex-D:
Maybe.

hurricane766:
Looks like this is a duplicate issue [#290](https://github.com/Alex-D/Trumbowyg/issues/290)

PR #445, opened 2016-11-17
add an option to fix android bug composition

### LINKED ISSUE #290: Key Code 229 prevents to trigger tbwchange
### Informations

Hi,

To start with, I am really enjoying the use of Trumbowyg! 

Now I know that from reading some of the comments on other issues logged and the github pages that mobile is not supported. Though it's not supported, from all the WYSIWYG, this is literally the best one out there (without paying an arm and a leg) for great simply editing on mobile!

With that being said I have small issue, on Android Browsers It looks like Android browsers were considered in the past, as at this part of the [code](https://github.com/Alex-D/Trumbowyg/blob/master/src/trumbowyg.js#L530), there are checks done agains the even code for key 229, which is a dreaded issues and long outstanding with [google](https://bugs.chromium.org/p/chromium/issues/detail?id=118639). I was wondering if we would be able to either remove the check or add a check against the browser. In short right now on line [530](https://github.com/Alex-D/Trumbowyg/blob/master/src/trumbowyg.js#L530) it gets checked and on line [550](https://github.com/Alex-D/Trumbowyg/blob/master/src/trumbowyg.js#L550) it prevents that key from being able to trigger a `tbwchange`

**Browser + version**:
Chrome 50.0.266
**OS**:
Android 6.0.1
**Resolution**:
### How to reproduce the bug?
- Open the demo page http://alex-d.github.io/Trumbowyg/ in a Chrome browser on Andriod
- Enter new text
- Observe that `tbwchange` never gets reached.

danieloprado:
Add an option to disable the composition check.
fixes #290 
```js
$('textarea').trumbowyg({
  useComposition: false
});
```

Maybe we can in future move this option to the language file.

Open this codepen on an Android mobile device:
http://codepen.io/danieloprado/pen/ObbXdv/

Issue related: 
Key Code 229 prevents to trigger tbwchange #290
What is the composition variable for? (trumbowyg.js:542) #411

### DIFF
--- a/src/trumbowyg.js
+++ b/src/trumbowyg.js
@@ -337,6 +337,7 @@ jQuery.trumbowyg = {
         // Defaults Options
         t.o = $.extend(true, {}, {
             lang: 'en',
+            useComposition: true,
 
             fixedBtnPane: false,
             fixedFullWidth: false,
@@ -548,7 +549,7 @@ jQuery.trumbowyg = {
             t.$ed
                 .on('dblclick', 'img', t.o.imgDblClickHandler)
                 .on('keydown', function (e) {
-                    composition = (e.which === 229);
+                    composition = t.o.useComposition && (e.which === 229);
 
                     if (e.ctrlKey) {
                         ctrl = true;


danieloprado:
Any updates?

Alex-D:
Sorry, your code is OK, thanks :)

PR #453, opened 2016-12-13
Add better support of keycode & composition for Chrome Android

JSteunou:
* by trying to catch to correct key code
* by not setting composition to true if the key code 229 is because of "crazy buffering keyboard mobile"
* by separating input from keyup events (maybe an option to not use input would be interesting, that would improve performances as input event is not necessary)

### DIFF
--- a/src/trumbowyg.js
+++ b/src/trumbowyg.js
@@ -544,12 +544,13 @@ jQuery.trumbowyg = {
 
             var ctrl = false,
                 composition = false,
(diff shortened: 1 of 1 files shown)


Alex-D:
Thanks! I will take a look ASAP.

JSteunou:
No problem. The root problem remain though, I still dont know why Chrome Android act like this. A simple `<div contenteditable="true"></div>` does not have this issue, keyboard is fine, so it might be something running wild inside your code.

Alex-D:
Thanks for this big fix :)

PR #457, opened 2016-12-16
Added select option for input type on modal

alcmoraes:
Hi Alex.

I was using Trumbowyg and I've missed the feature of creating a select inside a modal so I've designed this workaround.

In order to test it, just use:

    trumbowyg.openModalInsert(
        trumbowyg.lang.affiliate,
        {
            "Type": {
                type: 'select',
                options: {
                    'home': 'Home page',
                    'registration': 'Registration page',
                    'new_customer_bonus': 'New Customer bonus',
                    'special bonus': 'Bonus page'
                 },
                 required: true
         }
     }

And it works like a charm for now. But of course. As I said. Is just a workaround that I was in need now. Maybe someone else also needs this kind of changes.

Thanks and keep up the great work.

Cheers,

### DIFF
--- a/src/trumbowyg.js
+++ b/src/trumbowyg.js
@@ -1354,9 +1354,23 @@ jQuery.trumbowyg = {
                     return prop + '="' + a[prop] + '"';
                 }).join(' ');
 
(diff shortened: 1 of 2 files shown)


Alex-D:
Was too strong to be added to the core. And it's too old now. Sorry :/

Issue #323, opened 2016-07-10
Add an option for inserting smiley markup


AttitudeMonger:
I want to give an additional button for inserting smileys. When user clicks the button, a popup is shown nearby displaying some 30 smiley icons, clicking any of which inserts that into the caret position.

How do I add this button and the functionality here?

Alex-D:
Idk. Maybe on the color plugin base, changing colors squares by smileys and by calling `insertImage` with the smiley URL.

graphek:
any progress on this? i'm interested as well!

AttitudeMonger:
Not sure right now. I am planning to do it like Facebook, let the users use markup like :) or :( in the text, then at the time of getting the HTML and pasting ti back, parse through such markups, convert them to respective inline images and render alongwith the rest of the content.

graphek:
Sounds good! I'm currently looking in a implemenation with at.js

NicolasPion:
Hi,
I do this (based on colors plugin + trumbowyg.execCmd('insertText' ....), but i dont know if i can do a pull-request because it depends on [emojifyjs](https://github.com/Ranks/emojify.js) :

![screenflow](https://cloud.githubusercontent.com/assets/11868633/17366836/8b4e5fac-598d-11e6-94b0-ee598e0a7a97.gif)

graphek:
@NicolasPion 
wow, thats an awesome implementation! It would be really great, if you could share the code somehow :) I think that would be a great enhancement for the trumbowyg editor, emojis is the only thing i missed on the editor.

best regards

NicolasPion:
@graphek, thanks !
PR & files : https://github.com/Alex-D/Trumbowyg/pull/354
;-)

Alex-D:
You can PR them in two plugins :)

muyiwaoyeniyi:
@NicolasPion There is an issue with the scroll bar in the emoji dropdown.

Testing in:
Chrome 54.0.2840 and Latest Firefox
Ubuntu 16.04

Using version 2.3.0 but I replicated this issue in 2.4.2

You can't click on the emoji scroll bar. When you do so, the emoji dropdown goes away. I believe this is happening because the dropdown as a whole has a close on click event attached to it. But that affects the scroll bar. The emojis themselves probably need a container separate from the dropdown container.

I tried to track down where this is done in you code but I couldn't. If you point me in the right direction, maybe I can fix it. Thanks.

Thanks.

[...]

PR #459, opened 2016-12-16
Prevent dropdown is hidden when click on scrollbar

NicolasPion:
Resolve issue https://github.com/Alex-D/Trumbowyg/issues/323#issuecomment-267604714


Alex-D:
Thanks for this fix :)

PR #465, opened 2016-12-19
Limit emoticon size in Emoji plugin

NicolasPion:
Newest version of emojify.js provide biggest icons, so we need to limit their sizes in the editor


Alex-D:
I trust you!

PR #467, opened 2016-12-30
Remove the version number of VuejS

Elhebert:
I finally had the time to create a VueJS 2 implementation.

In the repo readme I added a link to tue VueJS 1 directive, thus, we can remove the version number from the gh-page.


Alex-D:
Yeah. GG. 
Thanks:

PR #475, opened 2017-01-19
modify keyup event handling for IME input

hirokts:
When the input by Input Method Editor is fixed, copying to original textarea fails.
(Chrome:54.0.2840.90, MOZC, Ubuntu16.04LTS)
This makes unable to input and save with only japanese words.

It is because of e.which(e.keyCode) when input by IME (japanese) is fixed.

* FireFox  
  keydown:－, keyup:13
* IE  
  keydown:229, keyup:13  
* Chrome  
  keydown229, keyup:－
* safari  
  keydown:229, keyup:13


Alex-D:
I trust you, I can't test.
Thanks! :)

hirokts:
Thanks!

PR #483, opened 2017-01-26
modify selection range control for IME input

hirokts:
When the input by Input Method Editor (IME) is confirmed, copying text to original textarea is succeed.
(previous pull request https://github.com/Alex-D/Trumbowyg/pull/475)

IME confirmation is also happens when a new character is entered after the conversion candidate is displayed with the space key.
But after the confirmation, new character will be reset by removing selection range because the character edited by IME  belongs to selection.


Alex-D:
I trust you, can't test IME.
Thanks for your time! :)

hirokts:
Thank you so much.

PR #489, opened 2017-02-08
added option textInParagraphs

brainfoolong:
Option to be able to disable automatic `<p>` paragraphs.
Doesn't change the current default behaviour but add more flexibility.
Also minor code format fixes (double/none space between operators)

### DIFF
--- a/src/trumbowyg.js
+++ b/src/trumbowyg.js
@@ -352,6 +352,7 @@ jQuery.trumbowyg = {
             semantic: true,
             resetCss: false,
             removeformatPasted: false,
+            textInParagraphs: true,
             tagsToRemove: [],
 
             btnsGrps: {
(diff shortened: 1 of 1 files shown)


src/trumbowyg.js
```diff
@@ -455,7 +456,7 @@ jQuery.trumbowyg = {
             try {
                 // Disable image resize, try-catch for old IE
                 t.doc.execCommand('enableObjectResizing', false, false);
-                t.doc.execCommand('defaultParagraphSeparator', false, 'p');
+                t.doc.execCommand('defaultParagraphSeparator', false, t.o.textInParagraphs ? 'p' : '');
```
Alex-D: I'm not sure that line works well in all browsers. Did you test it?

brainfoolong: Honestly, no. Cross browser support, as far as the documentations suggest, is for `div` and `p` only. So i guess this feature is also not for you because `div` tags are no option for trumbowyg, as stated in my other pull request.

Mohamm6d:
There is no way for this. Just make sure you delete all P tags when they are saving

PR #490, opened 2017-02-08
added new btnsDef for normal, div wrap, format

brainfoolong:
To be able to format a text without any special format (p, header, blockquote), i've added just another format option to wrap text with `div`, called 'normal' format, most editors handle it the same, or similar, way.

Default behaviour has changed, 'normal' btn is available by default.

Todo: Adding a svg icon for `div`.


brainfoolong:
Sry, wrong pull request, contains more changes, i'll close this and reopen it clean.

PR #491, opened 2017-02-08
added new btnsDef for normal, div wrap, format

brainfoolong:
To be able to format a text without any special format (p, header, blockquote), i've added just another format option to wrap text with div, called 'normal' format, most editors handle it the same, or similar, way.

Default behaviour has changed, 'normal' btn is available by default.

Todo: Adding a svg icon for div.

### DIFF
--- a/src/trumbowyg.js
+++ b/src/trumbowyg.js
@@ -7,6 +7,7 @@ jQuery.trumbowyg = {
             redo: 'Redo',
 
             formatting: 'Formatting',
+            div: 'Normal',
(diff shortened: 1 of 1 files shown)


Alex-D:
It can bug with semantic mode I think. So I can't merge that.

Also, div are not semantics. Il a blog post for example you should not have divs. Div are for structure, not for content.

You can add div by yourself with your own default configuration if you want. But I don't want that natively in Trumbowyg :/

PR #500, opened 2017-02-23
Get first tag when updating button pane status

Val-Git:
For block elements, such as the blockquote, the status of the buttons in the toolbar is not updated, if the element is empty because is always taken the parent element.


Alex-D:
Thanks for this fix :)

PR #509, opened 2017-03-01
Empty textarea value if editor contains only HTML elements

dawidchyrzynski:
If editor contains only HTML elements (without text nodes) syncTextarea does not update content of the textarea.


Alex-D:
Which kind of tag did you use?

Your "fix" will cause a bug: if I type a character, then delete it, the textarea will keep a `<p></p>` which cause the textarea is not empty. So if you want validate the textarea with a `required` attribute, it bugs :/

So you shoud prefer add your tags to the white list at the same line you change :)

PR #510, opened 2017-03-02
Update Japanese translation

chitoku-k:
The Japanese maintainer seems to be no longer paying attention to this project 😿


Alex-D:
Yay! Thanks :)

PR #534, opened 2017-04-15
Added .idea and the bower_components folders to .npmignore

ozdemirburak:
I think you have something like `core.excludesfile` config for `.idea` and `bower_components`, however, both of them are still included when someone downloads through the npm. 

But, if you are keeping them on purpose, then sorry for the PR :)


Alex-D:
You right thanks :) Will be fixed in next release!

PR #543, opened 2017-05-11
Correct lang

mrdr00py:
The language should be: jQuery.trumbowyg.langs.ro instead of jQuery.trumbowyg.langs.pt


Alex-D:
You right! Thanks! :)

PR #546, opened 2017-05-15
hyperlink anything

sinr0202:
After reviewing the document editor, I noticed links cannot include anything more than texts. this patch grabs the raw html to put inside the anchor tag.


Alex-D:
I don't think that works in any browsers :/

Alex-D:
Thanks!

I've check some more documentation and seems to be OK :)

sinr0202:
awesome! thanks for following up

PR #548, opened 2017-05-16
Allow custom emoji and fix emoji list

JaKhris:
1. This allows users to call custom emoji when instantiating Trumbowyg using an array instead of a string. (line 929-954)
2. This fixes a bug when the entire default emoji list is loaded in addition to a custom list of emoji. (line 915)

**Warning:** The images are added outside of a ```<p>``` tag and produce unexpected results. Ideas to fix this? @Alex-D

### DIFF
--- a/plugins/emoji/trumbowyg.emoji.js
+++ b/plugins/emoji/trumbowyg.emoji.js
@@ -912,7 +912,7 @@
         plugins: {
             emoji: {
                 init: function (trumbowyg) {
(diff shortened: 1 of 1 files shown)


Alex-D:
@NicolasPion can you review that PR please? :)

NicolasPion:
Works well !
Nice PR, and this remove "emojify.js" dependency (maybe we can use only this method ?!)

Alex-D:
Thanks! :)

PR #554, opened 2017-05-21
Added some needed basic security/filtering to the upload plugin.

mrwigster:
It's not incredible - but better than spreading the completely insecure file at current.


Alex-D:
I do not provide a backend script, it's juste a script for test purposes.

Alex-D:
See: https://github.com/mrwigster/Trumbowyg/blob/ce44ac8dd230e80e19503e63408a4a3ad2bb71a9/plugins/upload/trumbowyg.upload.php#L11

mrwigster:
That's fair enough - I just thought it may have been useful to share a script that was at least a tiny bit securer rather than the current.

Alex-D:
That's just for the dev phase, not for production, if the code is too long, it's hard to be efficient on the Trumbowyg side.

PR #555, opened 2017-05-22
CORS Correction

cleoo:
Allow svg xhr loading from cross domain (with Access-Control-Allow-Origin * for svg configured on server)


Alex-D:
Thanks! :)

PR #563, opened 2017-06-01
Fix for issue 557: Chrome Warning about Selection.addRange()

hello-monchi:
createLink,unlink: removeAllRanges on document selection with existing range; because updating a existing range is not allowed anymore (see: https://github.com/w3c/selection-api/issues/80 and https://www.chromestatus.com/features/6680566019653632)


Alex-D:
Thanks for the fix!

Issue #559, opened 2017-05-30
<script> being executed before being removed with tagsToRemove


ffaubry:
When using 'tagsToRemove' with 'script', the script is still executed before being removed.

- use 'viewHtml'
- copy paste: <script>alert('hello');</script>
- switch back to normal editor (not htmlView)

You will see that the code is executed. This is a security hole in my opinion.

I don't think that #160 was correctly fixed.

We will submit a PR to fix this.

Alex-D:
I do not want fix that.
Again, it's to your backend to secure that. Before that, only the user itself is annoyed.

ffaubry:
Alex,
The issue is that users don't know what they copy/paste. 
When they switch the context from viewHtml to not viewHtml, the scrip is executed and malicious things can happen. At that moment it is not just an issue of being annoyed. Bad things could have happened under the domain that is hosting the app.
I'm asking you to reconsider this issue.
I speak french too. We can have a phone call if you want.
This a serious issue. We really want to make Trumbowyg safe.

Alex-D:
Ok, so make a PR and I will check that :)

ffaubry:
@Alex-D Thanks for considering the PR we just submitted.
This is a big deal us. Thanks in advance.

PR #565, opened 2017-06-05
Fix for issue #559

ffaubry:
Related issue: #559 

Issue: 
Trumbowyg editor is loading and executing tags while toggling between HTML and text view that are marked for removal in the tagsToRemove

Solution:
The synCode function takes the HTML from the HTML view and loads it into the Text View. The synCode loading the html which will execute the code the solution is to remove the tags before loading into the doc. Since the synCode function is called from several other functions it’s ideal place to remove the tags right before loading into the doc.

It is using jQuery to load the html and use the remove method to remove the tags the other option is use regular expression to remove the tags but found out regular expression is not fool proof and could be exploited by making a complicated html.

The syncTextarea  function is taking the text from the Text view and loading it into the HTML view. Since the html is encoded it’s safe and the tags won’t be executed.

This has been tested in our environment by our QA team and all the security issues have been resolved.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,6 +1,6 @@
 {
   "name": "trumbowyg",
-  "version": "2.5.1",
+  "version": "2.5.2",
   "homepage": "https://github.com/Alex-D/Trumbowyg",
   "authors": [
     {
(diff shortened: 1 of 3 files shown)


ffaubry:
@Alex-D Thanks a lot!

Issue #567, opened 2017-06-06
$.trumbowyg('html', '...new content..') No longer works after 2.6.0


robinfhu:
### Informations

**Browser + version**: Chrome latest
**OS**: Mac OS
**Resolution**:  Default

### How to reproduce the bug?
**Using Trumbowyg version 2.6.0:**

If you try to do `$('textarea').trumbowyg('html',"Hello World");`, it will not update the text area with the new content.

This used to work in older versions.

fdelayen:
Same here, seems like this part of the api is broken :/

arabsight:
yeah, same here.

benpoulson:
Same issue here. Works if you roll back to `2.5.1`.

Alex-D:
I've checked the diff, I did not understand what's the change, if you could investigate too, it could be cool :)

wygodad:
Did the problem have been resolved?

benpoulson:
@Alex-D I can't spot any obvious causes. I assume it's something to do with lines 1017 [here](https://github.com/Alex-D/Trumbowyg/compare/v2.5.1...v2.6.0#diff-a57e0e52aed1451208c5f2010cd3b537).

My issue does involve the WYSIWYG being hidden when instantiated, and the code here does check for visibility beforehand.

I'll have a play about with the source this evening.

OlivierB-OB:
Hi @Alex-D ,

I managed to fixthis issue.
I made a pull request with my fix proposal: [https://github.com/Alex-D/Trumbowyg/pull/569](https://github.com/Alex-D/Trumbowyg/pull/569)

Rgds,
Olivier

richardtallent:
An hour of my life wasted thinking this was a problem in wiring up Vue properties... :(

srinudhulipalla:
I am also facing the same issue. Is there a way to fix this 2.6 or shall we go back to 2.5?

srinudhulipalla:
I went back to 2.5.1 version. I works great..

OlivierB-OB:
Hello @Alex-D  !

Did you find time to have a look at my fix proposal?
https://github.com/Alex-D/Trumbowyg/pull/569

OlivierB-OB:
Thanks a lot!

PR #569, opened 2017-06-14
#567 - $.trumbowyg('html', '...new content..') No longer works after …

OlivierB-OB:
…2.6.0


Alex-D:
Thanks! It works well :)

OlivierB-OB:
Thanks a lot!

PR #590, opened 2017-07-10
Paste event on textarea

nurogenic:
Paste event on textarea fires `tbwchange` event before paste content is on element.


Alex-D:
Thanks :)

PR #597, opened 2017-07-20
Fix keyboard shortcuts not working on macOS

boylett:
Always check for both ctrlKey and metaKey when using keybindings 👍


Alex-D:
Should we only listen for metaKey on mac?

boylett:
Personally, I think it's better to cover both as mac's System Preferences allows users to control which key counts as Command or Control, so some may be using CTRL for most of their meta-tasks.

screen shot 2017-07-21 at 15 10 47

PR #609, opened 2017-08-11
Replaced `height` and `top` calculation with CSS

marcinwieprzkowicz:
Hi @Alex-D,

I really like Trumbowyg WYSIWYG editor. I had a pleasure to look through the source code and I found that you are calculating lots of properties with JavaScript, where you can easily replace it with CSS.

Please take a look at code in this PR. Easy replacement for `height` and `top` in overlay element.

Cheers


Alex-D:
Thanks! :)

PR #620, opened 2017-09-06
Update trumbowyg.js

JoSe086:
When the enter key is pressed in chrome, firefox and edge, the (keyCode == 13) is false and a false value is passed to the t.semanticCode function (and the enter key works fine).

When the enter key is pressed in IE browsers, the (keyCode == 13) is true and a true value is passed to the t.semanticCode function. Pass true to this function, causes that the editor can't create new line breaks.

If we always pass false to this function, the enter key seems to works fine on all browsers.

### DIFF
--- a/src/trumbowyg.js
+++ b/src/trumbowyg.js
@@ -600,7 +600,7 @@ jQuery.trumbowyg = {
                     if ((e.ctrlKey || e.metaKey) && (keyCode === 89 || keyCode === 90)) {
                         t.$c.trigger('tbwchange');
                     } else if (!ctrl && keyCode !== 17) {
-                        t.semanticCode(false, keyCode === 13);
+                        t.semanticCode(false, false);
                         t.$c.trigger('tbwchange');
                     } else if (typeof e.which === 'undefined') {
                         t.semanticCode(false, false, true);


src/trumbowyg.js
```diff
@@ -600,7 +600,7 @@ jQuery.trumbowyg = {
                     if ((e.ctrlKey || e.metaKey) && (keyCode === 89 || keyCode === 90)) {
                         t.$c.trigger('tbwchange');
                     } else if (!ctrl && keyCode !== 17) {
-                        t.semanticCode(false, keyCode === 13);
+                        t.semanticCode(false, false);
```
Alex-D: Should be just `t.semanticCode()` since false are defaults

Alex-D:
I've make a better fix after some tests :)

JoSe086:
Nice!!!

The changes will be merged on 2.8.0 branch or can I download the develop
branch to be used as production code?

Alex-D:
You should use 2.7.2 for now:

- use npm: `npm install trumbowyg`
- use yarn: `yarn install trumbowyg`
- use bower: `bower install trumbowyg`
- https://github.com/Alex-D/Trumbowyg/archive/master.zip

2.8.0 is WIP for now, not stable.

PR #632, opened 2017-09-25
Ruby text plugin?

0x00000F5C:


Alex-D:
I did not understand the goal of this plugin. I did not know what you means by "ruby text".

0x00000F5C:
Ruby text is phonetic guide; smaller letters above a text to indicate how it reads in another language, or plain alternative reading. I needed it since I'm dealing with Japanese characters in my project. I don't know, someone might need it.

It's something like [this.](https://www.w3.org/International/articles/ruby/)

PR #644, opened 2017-10-18
add rtl option to persian language

### LINKED ISSUE #643: persian like arabic need _dir: 'rtl'
### Informations

Persian is an rtl language and its need _dir: 'rtl' in language options

dariubs:
fix #643


Alex-D:
Thanks :)

Issue #566, opened 2017-06-06
There is no form change if you delete content with Ctrl + A and Delete


FiloSpaTeam:
Before report an issue, be sure to put these informations. Thanks :)

### Informations

**Browser + version**: Chromium - 58.0.3029.110 Fedora Project (64-bit)
**OS**: Fedora 25 64bit
**Resolution**: Monitor?

### How to reproduce the bug?
If you write content into textarea (with trumbowyg) and later you try to delete with Ctrl + A (select all) and Delete and you click out, no change event is propagated.

vetruvet:
I'm running into the same exact issue. This also happens if you have just one word in the input and you use Ctrl+Backspace/Ctrl+Delete to delete it. Also change events stop for a bit (200ms to be precise) if you are typing and just press the Ctrl key.

I dug through the code and I can shed some light on this issue. I don't have a good solution but it may help @Alex-D find/address the issue.
There is a variable that tracks when the Ctrl key is pressed. It's set to true [here](https://github.com/Alex-D/Trumbowyg/issues/566). There is then a 200ms timer set [here](https://github.com/Alex-D/Trumbowyg/blob/develop/src/trumbowyg.js#L609) to reset it to false, but that timer is _only_ fired when there some input/change happens (which it does not on Ctrl+A for example). Then [here](https://github.com/Alex-D/Trumbowyg/blob/develop/src/trumbowyg.js#L602) is the check that prevents the change event from firing when the Ctrl key is pressed down.
I'm not sure about the original intention of this code, hence why I'm not offering any solutions now

kmarcisz:
This issue stems from fix applied in https://github.com/Alex-D/Trumbowyg/issues/311 which could not be reproduced back then. If possible I say get rid of the debounce, or at the very least make it optional via trumbowyg config.

Meekohi:
Having this issue as well.

vetruvet:
@Meekohi good catch - updated the links to point to what is the latest commit on the develop branch as of now.

Meekohi:
Another way this bug manifests: if you press and release `Ctrl` then type a letter -- this also does not trigger `tbwchange`.

Meekohi:
Added a pull request that should fix this. I might be misunderstanding the intent of the `ctrl` variable here, but I assume it should "reset" once someone is no longer holding down that key.

Alex-D:
This should be fixed by @Meekohi's fix :)

PR #650, opened 2017-10-30
Fix issue #566

Meekohi:
This fixes Issue #566 by resetting `ctrl` on `keyup` in 200ms (if `Ctrl`/`Meta` isn't still being held down, and `Alt` isn't being held down)


Alex-D:
Thanks! :)

PR #654, opened 2017-11-07
Increase z-index to 12

Shonetow:
With z-index 11 the dropdown-menu goes under the toolbar of the next editor instance.

image


Alex-D:
Good catch! Thanks :)

PR #659, opened 2017-11-16
Prevent open dropdown menu when disabled editor

ne0h12:
Hi.
When editor is disabled, but dropdown buttons still work.


Alex-D:
I've really fixed the bug globally :)

Thanks for the report!

PR #663, opened 2017-11-24
Allow users to set the width of an image in the modal

AbdelElrafa:
We should allow them to edit the width of an image easily in the modal.


AbdelElrafa:
@Alex-D I have added this change to my code and it really is a nice feature. I saw before in many comments you rejected the idea but I think that this solution is clean and doesn't hurt to have it. Most people aren't smart enough to adjust the size of their images locally before pasting.

Alex-D:
OK. But this should be an option set to false by default.

AbdelElrafa:
@Alex-D Done. Should I add height too? I don't think it would be a good idea to since that can cause distortion.

Alex-D:
It looks good to me :)

PR #672, opened 2017-12-08
[ allowtagsfrompaste ] Fix issue with removing the wrong tags

edwinveldhuizen:
if "s" is not in the allowTagsFromPaste list, then "<span>foo</span>" is removed as well with the current regex.

This commit is preventing that from happening

### DIFF
--- a/plugins/allowtagsfrompaste/trumbowyg.allowtagsfrompaste.js
+++ b/plugins/allowtagsfrompaste/trumbowyg.allowtagsfrompaste.js
@@ -133,7 +133,7 @@
 						setTimeout(function () {
 							var processNodes = trumbowyg.$ed.html();
 							$.each(trumbowyg.o.plugins.allowTagsFromPaste, function(iterator, tagName) {
-								processNodes = processNodes.replace(new RegExp("<\\/?" + tagName + "(\\s[^>]*|)>", "gi"), "");
+								processNodes = processNodes.replace(new RegExp("<\\/?" + tagName + "(\\s[^>]*)?>", "gi"), "");
 							});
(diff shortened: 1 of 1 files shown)


plugins/allowtagsfrompaste/trumbowyg.allowtagsfrompaste.js
```diff
@@ -133,7 +133,7 @@
 						setTimeout(function () {
 							var processNodes = trumbowyg.$ed.html();
 							$.each(trumbowyg.o.plugins.allowTagsFromPaste, function(iterator, tagName) {
-								processNodes = processNodes.replace(new RegExp("</?" + tagName + "[^>]*>", "gi"), "");
+								processNodes = processNodes.replace(new RegExp("<\\/?" + tagName + "(\\s[^>]*|)>", "gi"), "");
```
Alex-D: `"<\\/?" + tagName + "(\\s[^>]*|)>"`

Should be: 

`"<\\/?" + tagName + "(\\s[^>]*)?>"`

No?

edwinveldhuizen: No, it looks funny, but it either ends with space + params or nothing at all:
<span foo=“bar”>
</span>

edwinveldhuizen: Ah actually `"<\\/?" + tagName + "(\\s[^>]*)?>` works as well indeed, 
guess it doesn't really matter

Alex-D:
Thanks :)

PR #673, opened 2017-12-11
Reduce ctrl button lock-in to 50ms

Meekohi:
Related to Issue #566, I'm still finding that some users trying to do `Ctrl+A`, `Delete` are not seeing their change propagate because of the 200ms lock-in after releasing the Ctrl button. Recommend changing this to 50ms to give a better chance of this working.

I'm not sure what the origin of the lock-in is exactly. If someone knows where that came from maybe we can find a solution that works for both use cases. I'm guessing this was just to avoid some corner case where people release the `Ctrl` key a little too early?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "trumbowyg",
-  "version": "2.7.0",
+  "version": "2.8.1",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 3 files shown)


Alex-D:
I think it was before composition and ctrl was splited.

So it's seems legit to reduce that duration.

Alex-D:
Thanks :)

Meekohi:
Sounds good, although if the timeout/lockin doesn't serve a purpose anymore should I just remove it entirely?

Alex-D:
Idk really... we need to test it in different cases.

Issue #618, opened 2017-09-06
Cleanpaste plugin window freeze


vsindylek:
### Informations

**Browser version**: Chrome 60.0.3112.113
**OS**: Win 10
**Screen resolution**: desktop (irrelevant)
**Trumbowyg version**: 2.7.0

### How to reproduce the bug?
Activate cleanpaste plugin and paste whatever into editor. It causes me to completely freeze the browser tab and wouldn't do anything - stops completely reacting and I found no way to debug it because DevTools freezes also.

arturseletskiy:
+1, but in source code mode it works well

franzos:
+1, same issue

**Browser version**: Version 61.0.3163.100
**Trumbowyg version**: 2.8.0

slackwalker:
It appears that the root cause of this is that the content of the text editor is changed by the paste before any of the pasteHandlers are called. The code in cleanpaste that's used to detect what was pasted into the text editor can't work because it plugins don't have access to what the text was before the paste occurred. Perhaps difference detection should be removed and any time a paste occurs, all text in the editor should be cleaned?

My pull request does fix the freeze, but doesn't solve the root issue, and ends up pasting three copies of clipboard text whenever a paste event occurs (not ideal).

Alex-D:
Should not freeze with the fix merged and published in 2.9.3 :)

Alex-D:
@slackwalker 
> Perhaps difference detection should be removed and any time a paste occurs, all text in the editor should be cleaned?

Can you detail more about that?

slackwalker:
Sure.

The cleanpaste plugin sets up a pasteHandler that attempts to grab the text content of the editor before the paste event occurs. Then it makes a setTimeout call, presumably to get the content of the editor after paste processing has completed. It then goes through a painful difference detection process with the contentBefore, and contentAfter text. The idea I think is to only clean up the newly pasted text.

The problem is that Trumbowyg has already changed the content of the editor before it hands off to any custom pasteHandlers from plugins so contentBefore and contentAfter will always be the same (unless another pasteHandler plugin made changes, but let's not pull that thread for now.)

Since this plugin is just cleaning text on paste, and cleaning it twice really shouldn't be a problem, maybe it would be more performant and functional to just run the cleaning code across the entire contents of the editor every time a paste occurs. This is how I've modified the plugin locally to work. It really guts the plugin though.

I can setup a pull request for the version of the plugin I'm running if you like.

[...]

PR #676, opened 2017-12-14
Fix Cleanpaste plugin window freeze #618

slackwalker:
Create minLength var that stores the shortest length of htmlBefore and htmlAfter, and ends comparison at that length if no differences are discovered by then. charAt will return an empty string if addressing beyond the end of a string, which can create an endless loop in this case.

Fix adds implicit dependency on javascript Math lib, and creates two lines that exceed 80 characters in width.


Alex-D:
Sorry for the long time :/

It seems to be OK :)

Thanks!

PR #678, opened 2017-12-17
Add align feature to insertImage

kzvonov:
Hi, I want to suggest a small improvement, add `align` attribute to images.
This feature will be accessible by the flag `imageAlignModalEdit`.
```
$('#text').trumbowyg({
    imageAlignModalEdit: true
});
```

If I did something wrong, please tell me and I will fix)

And thank you for great editor)


Alex-D:
Sorry for the long time. I'm not sure about this modification since it's not a select and it's not really user-friendly. Why did you not edit the HTML at this time?

Issue #671, opened 2017-12-07
Upload Plugin: Image Size


yongjiajie:
### Informations

**Browser version**: Chrome Version 62.0.3202.94 (Official Build) (64-bit)
**OS**: ubuntu 16.04 LTS
**Screen resolution**: 1920 x 1080

### How to reproduce the bug?
Not a bug.

After uploading an image in Trumbowyg via the Upload plugin, is it possible to control the size of the image uploaded?

Alex-D:
What did you mean by "size"? If it's width, user can set a width.
If you want put some width/height, you should use some CSS on img tag.

yongjiajie:
Hello @Alex-D, I am interested in showing thumbnails of the image in Trumbowyg, as opposed to the entire image uploaded. Thank you for your reply.

Alex-D:
You just had to send the URL of the resized thumbnail image, Trumbowyg do not care about the original image size :)

Issue #168, opened 2015-08-16
Image size adjusting


gaaarfild:
When uploaded new image and trying to adjust size, it sets correctly in CSS, but image render stays as it is.

Alex-D:
I don't understand, can you post a screencast or a example?

gaaarfild:
http://youtu.be/lLLh90zI6bI

Alex-D:
I think your CSS has a think like `img { width: 100% }`

gaaarfild:
I found a line in trumbowyg.min.css
`width: auto;`
```
.trumbowyg-editor embed, .trumbowyg-editor img, .trumbowyg-editor object, .trumbowyg-editor video
```
for this selector. If I remove it, everything become correct.

llogiq:
Can confirm this happens on current firefox with the demonstration page.

Ydalb:
Hm, sorry to re-open it, but how do you get the little square to resize the picture ? I don't have it :(
If I double click on the picture, it opens the modal to edit URL & description.
 
![capture d ecran 2017-04-05 a 16 20 26](https://cloud.githubusercontent.com/assets/1888805/24709934/d15968a8-1a1b-11e7-80ba-241144857ccd.png)

Issue #157, opened 2015-07-13
Can't resize the image


klocus:
On Opera (30) and IE (11), I can't change the image size. I don't have even squares on corners of the image.

Alex-D:
You need to upload the image at the good size to avoid users transform wallpapers in icons. That's the Trumbowyg philosophy.

adon988:
OMG

Issue #142, opened 2015-06-14
Image resize bug


Samuell95:
Hello,
i find image resize bug. If i resize image in editor no change to image size.
In chrome resize border doesn't work.

I find if `width="100" height="100"` change to css `style="width:100px;height:100px"` it work and resize it

Alex-D:
The editor don't support image resize. You need to upload the image with the good image. It's a best practice to optimize requests.

PR #680, opened 2017-12-22
new plugin 'resizimg': add image resizing capability

geonanorch:
I understand that image resizing can be a bad thing when huge files are downloaded only to display a small 32x32 picture in the browser. So OK no resizing in Trumbowyg Core.

But clearly (looking at #14, #142, #157, #168, and more recently #671 (and I probably missed a few issues) there is a need for that functionality in some cases. 

I implemented for my own needs a plugin (using another jQuery plugin) which will make all images resizable by dragging their bottom-right corner and though I would give back to the community by contributing --because Trumbowyg is about the best open editor around here :-)

Possible future work:
- relative size instead of absolute size
- resize handle (hopefully pure CSS)


Alex-D:
I'm not really OK to get `plugins/resizimg/jquery-resizable.min.js` into Trumbowyg repository.
It should be a link to a CDN/official site in docs :)

And for dev purpose, maybe add it to the package.json dev dependencies.

Alex-D:
What's up?

geonanorch:
Hi Alex-D, and thank you for your interest :-)
I understand, it does not make sense to include an external dependency in the code. Here is the situation: 

To support `resizimg` I had to add a `destroy` method to the `resizable` plugin. I submitted a [pull request](https://github.com/RickStrahl/jquery-resizable/pull/27) for that plugin, but haven't heard from @RickStrahl  until now.

Possibly he is still on holiday or busy. I propose to wait another week (and I will ping him). If nothing happens then I would merge the `resizable` code into `resizimg` and update this PR (I made a couple of improvements in the meantime anyway). OK for you ?

geonanorch:
OK, the `jquery-resizable` plugin was updated yesterday so I was able to externalize it.
I just pushed a new version of the `resizimg` plugin reflecting this and including other fixes/improvements.
I tested the live demo using a local HTTP server (the current demo framework does not seem to support external links though, I had to hack a bit), ready for you to review.

jung-hunsoo:
It would be great if I can use this feature in Trumbowyg.

Most of my webapp users are not good at using image editor, but a general end-user. And I'd like to provide as convenient as possible service.

Alex-D:
I just need to take the time to try it, then I will merge that :)

dartg:
Thanks a lot for this! Very useful.

afaraha8403:
Hey @Alex-D  & @geonanorch 

Just wondering about the resize handles.  Anything yet?

Thanks for the great work, I am loving this 👍

Alex-D:
Feel free to add them in a Pull Request, I think you can do that with an img::before (or ::after) + box-shadow hacky CSS way :)

geonanorch:
@afaraha8403 images are replaced elements, more difficult to manage than regular elements. I developed a solution for my needs, but it involves Javascript (+ jquery) and is a bit 'hacky'.
@Alex-D I would be interested to implement a pure CSS solution, can you describe the hack you have in mind? Afaik `before::` & `after::` are not good for replaced elements, but I do not know about that `box-shadow` trick, happy to learn!

geonanorch:
@Alex-D that is the replaced element issue I mentioned earlier, that approach does not work (basic background on [stackoverflow](https://stackoverflow.com/questions/7396469/why-dont-before-and-after-pseudo-elements-work-with-img-elements)). I solved it by adding a wrapper around IMG elements, but on top of being a bit convoluted it is not 100% transparent for contenteditables / your editor. Maybe you can think of something better?

Alex-D:
Arf OK.

So, what about an DOM element which will be manipulated in JS to change his absolute position to be over the image, but with `pointer-event: none`? With that, you do not change the content, only add an editor "meta" DOM element.

PR #686, opened 2018-01-06
Update trumbowyg.cleanpaste.js to clean all text #618

slackwalker:
Rewrite paste handler to clean all text in editor, instead of trying to determine which is newly pasted text first.

### DIFF
--- a/plugins/cleanpaste/trumbowyg.cleanpaste.js
+++ b/plugins/cleanpaste/trumbowyg.cleanpaste.js
@@ -13,14 +13,6 @@
 (function ($) {
     'use strict';
 
-    function reverse(sentString) {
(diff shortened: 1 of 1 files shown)


slackwalker:
I failed to get it in the commit message, but this references #618

edwinveldhuizen:
One great enhancement for this plugin would be to have options for which cleanups to handle and which to ignore.

For example, I already use the allowtagsfrompaste plugin with a custom list of allowed tags, so I don't want cleanPaste to do remove it's own list of tags.

Any chance you could include that?

slackwalker:
@edwinveldhuizen That's an interesting idea. I'll play around.

PR #688, opened 2018-01-09
fix: paragraph issue on first text element

### LINKED ISSUE #516: First paragraph isn't wrapped around with P tags
After updating to the latest version 2.5.1. When I first write a paragraph and press enter, the first paragraph isn't wrapped around with p tags, whether on the second ENTER, it does. It wasn't like that in the 2.4.2 version. In the older version, when I pressed ENTER after finishing the first paragraph, it was inside P tags.

Using latest Chrome browser Version 56.0.2924.87 on Windows 7.

alvesjtiago:
Fixes #516

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "trumbowyg",
-  "version": "2.9.2",
+  "version": "2.9.3",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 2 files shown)


Alex-D:
Thanks! :)

alvesjtiago:
Thank you for the great project 😊

paul-crashley:
Is this change in v2.10.0? I've just run into this issue myself using the latest tagged version.

Alex-D:
Yep, this commit is in v2.10.0

PR #701, opened 2018-01-24
Fix double slash in generated svgPath

jhdrn:
The previous implementation resulted in a double slash before "ui" in the path to the SVG icons, i.e. /foobar//ui/icons.svg.


Alex-D:
Thanks

PR #708, opened 2018-02-08
[ Wording ] Small updates

edwinveldhuizen:
They're just small, but the wording matches better with MS Word and are less "technical"

### DIFF
--- a/plugins/colors/trumbowyg.colors.js
+++ b/plugins/colors/trumbowyg.colors.js
@@ -19,8 +19,8 @@
                 backColor: 'Barva pozadí'
             },
             en: {
-                foreColor: 'Text color',
-                backColor: 'Background color'
+                foreColor: 'Font Color',
+                backColor: 'Background Color'
             },
             fr: {
                 foreColor: 'Couleur du texte',
--- a/src/trumbowyg.js
+++ b/src/trumbowyg.js
@@ -17,25 +17,25 @@ jQuery.trumbowyg = {
             strikethrough: 'Stroke',
             underline: 'Underline',
 
-            strong: 'Strong',
-            em: 'Emphasis',
+            strong: 'Bold',
+            em: 'Italic',
             del: 'Deleted',
 
             superscript: 'Superscript',
             subscript: 'Subscript',
 
-            unorderedList: 'Unordered list',
-            orderedList: 'Ordered list',
+            unorderedList: 'Bullets',
+            orderedList: 'Numbering',
 
             insertImage: 'Insert Image',
             link: 'Link',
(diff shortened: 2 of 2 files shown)


Alex-D:
I'm not OK to change that officially. If you want to change that in your project, you can by using your own language file.

Sorry :/

edwinveldhuizen:
Ok, but currently the wording is a bit inconsistent, so what is the process for changing existing wording?
Or can one only add new languages and then no-one can change it anymore? ever?

edwinveldhuizen:
'Strong', 'Emphasis', 'Unordered list' and 'Ordered list' are not really terms that are used

Alex-D:
> Or can one only add new languages and then no-one can change it anymore? ever?

I did not understand your point.

If you want change the wording in your project, you can, you just need to create your own language file, then use it.

https://alex-d.github.io/Trumbowyg/documentation/#add-localization

Alex-D:
Also, bold and strong are not the same. Since Trumbowyg try to be semantic, it support both.

Strong is the `strong` tag, bold is the `b` one. Which is not the same at all.

edwinveldhuizen:
I get your point from a technical perspective, but if we have a look at the following languages, they all call them the same as well.
The reason that we call them the same, is because for the end-user they are the same thing. Technically not the same at all, but for a less-technical end-user ( who are the actual users of a wysiwyg editor ) they are the same.

*bg*
```
    bold: 'Удебелен',
    italic: 'Наклонен',

    strong: 'Удебелен',
    em: 'Наклонен',
```

*by*
```
    bold: 'Паўтлусты',
    italic: 'Курсіў',

    strong: 'Паўтлусты',
    em: 'Курсіў',
```

*cs*
```
    bold: 'Tučné',
    italic: 'Kurzíva',

    strong: 'Tučné',
```

*el*
```
    bold: 'Έντονα',
    italic: 'Πλάγια',

    strong: 'Έντονα',
    em: 'Πλάγια',
```

*es*
```
    bold: 'Negrita',
    italic: 'Cursiva',

    strong: 'Negrita',
```

*fa*
```
    italic: 'مورب',
    em: 'مورب',
```

And many more

Alex-D:
OK. So make me a new PR but without capitalize.

edwinveldhuizen:
`Align Left',  'Align Right' and 'Insert Image' are currently also capitalised.
That's why we were making it consistent

edwinveldhuizen:
Shall I change those to non-capitalised as well, or keep all capitalised?

Alex-D:
Change to non-capitalised :)

Thanks for your time!

Issue #709, opened 2018-02-09
openModalInsert allows fields with a name attribute when building the html, but ignores it when interpreting the result


nikkilocke:
### Informations
**Browser version**: Google Chrome | 63.0.3239.132 (Official Build) (64-bit) (cohort: Stable)
**OS**: Windows 10
**Screen resolution**:1080x1920
### How to reproduce the bug?
I am trying to use radio buttons in an openModalInsert. I know this is not supported, but the bug applies to any field where a `name` attribute is set.

				imgDblClickHandler: function () {
					var t = i.data('trumbowyg'),
						$img = $(this),
						src = $img.attr('src'),
						base64 = '(Base64)',
						style = $img.attr('style') || '';

					if (src.indexOf('data:image') === 0) {
						src = base64;
					}

					var options = {
						url: {
							label: 'URL',
							value: src,
							required: true
						},
						alt: {
							label: t.lang.description,
							value: $img.attr('alt')
						},
						noFloat: {
							label: 'No Float',
							type: 'radio',
							name: 'floatImage',    // **** Same name to force radio group
							value: ''
						},
						floatLeft: {
							label: 'Float left',
							type: 'radio',
							name: 'floatImage',    // **** Same name to force radio group
							value: 'float:left'
						},
						floatRight: {
							label: 'Float image to right',
							type: 'radio',
							name: 'floatImage',    // **** Same name to force radio group
							value: 'float:right'
						}
					};
					if(style.search('left') >= 0)
						options.floatLeft.attributes = { checked: true };
					 else if(style.search('right') >= 0)
						options.floatRight.attributes = { checked: true };
					else
						options.noFloat.attributes = { checked: true };
					if (t.o.imageWidthModalEdit) {
						options.width = {
							value: $img.attr('width') ? $img.attr('width') : ''
						};
					}

					t.openModalInsert(t.lang.insertImage, options, function (v) {
					return false;
				}

openModalInsert has this code:

            $.each(fields, function (fieldName, field) {
                var l = field.label,
                    n = field.name || fieldName,    // **** Note - name attribute is explicitly catered for
                    a = field.attributes || {};

                var attr = Object.keys(a).map(function (prop) {
                    return prop + '="' + a[prop] + '"';
                }).join(' ');

[...]

PR #712, opened 2018-02-14
openModalInsert fixes and enhancements for issue 709

nikkilocke:
Fix bug where, if a name is provided which is different to the fieldName, an exception was thrown in the OK button handler.

When populating the results in values, use name (if provided) rather than the fieldName.

Explicitly support radio buttons (multiple type:'radio' options with the same name), returning the name element in values to the value of the selected radio button.

To compensate a little for the few extra lines of code, optimise the $.each which builds the dialog html slightly.

Add a little more documentation in docs/documentation/core/index/html and docs/documentation/index.html (I was not sure whether one was generated automatically from the other, so updated both).

*** NOTE ***

I have not updated any of the files in dist. I assume that is done as part of the release process.

[Issue 709](https://github.com/Alex-D/Trumbowyg/issues/709) refers.


Alex-D:
Looks good to me, but I need to test it :)

PR #714, opened 2018-02-15
Start fullscreen option

adalenv:


Alex-D:
I was thinking about an API to allow to exit fullscreen too :/

adalenv:
there is the button for that lol

Alex-D:
Yep but if you are a dev and want to open/close fullscreen the best is an API, not an option on the constructor :)

adalenv:
how will you do it?

Alex-D:
https://github.com/Alex-D/Trumbowyg/blob/develop/src/trumbowyg.js#L175

Something like that.

adalenv:
it will not open in fullscreen if you have to asign other options

PR #718, opened 2018-02-20
[Plugin] History plugin for enhanced undo/redo functionality

forelabs:
## History plugin

**Enhanced Undo / Redo functionality by watching changes to the editor**

history

### How does it work?

The plugin listens to `tbwinit` and `tbwchange`  events on `trumbowyg.$c` (instance wise)
and then check if the change is something new:

- If the focused node is a text node it watches for any whitespace character and will create a new stack entry if found, otherwise it will check if the focused node has changed since the previous change, if yes it will also create a new stack entry, otherwise it will exchange latest stack entry with newest one.

- If it is not a text node it compares the html tags within the editor of the new state with the previous state. the note types, the count and the order of the elements has to be the same.

- The last special rule is for the first stack entry, so if there will be the first change it will always create a new stack entry, so you can go back to the initial state.

### Ehanhanced undo / redo ?

There are big advantages to the default undo / redo buttons:

- it is not limited to browsers which support simple undo / redo
- it saves all changes to the editor, not only the simple ones (e.g. inserted something through a plugin)

*As you might noticed my table plugin is also enhanced. I know there is a open pull request to enhance the table plugin, saw this one too late and also I think my table plugin is a bit more user friendly. I think I will create another pull request for this one, but first I have to extract my icons to svgs i used here. If you want to have a short look, see here:* [enhanced table plugin](https://user-images.githubusercontent.com/3809644/36413141-67228b5a-161d-11e8-856b-4f5e69723b58.gif)


docs/documentation/plugins/index.html
```diff
@@ -273,16 +300,20 @@ <h3 id="plugin-insertaudio">Insert audio</h3>
 
             <h4>How to use it?</h4>
             <pre><code class="html">
-&lt;-- Import Trumbowyg plugins... -->
-&lt;script src="node_modules/trumbowyg/dist/plugins/insertaudio/trumbowyg.insertaudio.min.js">&lt;/script>
+&lt;-- Import Trumbowyg history style in &lt;head>... -->
```
forelabs: I have put those lines in the wrong section, I will fix that tomorrow.

Alex-D:
Seems really good!

I need to take time to review this.

And yeah, send this awesome table plugin update! :D #need

PR #722, opened 2018-02-26
[Fix] Table plugin scss missing bracket

forelabs:
I have somehow missed the last bracket on my PR, `gulp build` threw an error because of that.
Recognized due to checkout of the latest develop branch.


Alex-D:
Already fixed, no?

https://github.com/Alex-D/Trumbowyg/blob/develop/plugins/table/ui/sass/trumbowyg.table.scss

PR #725, opened 2018-02-27
[Feature] createLink read text from link

forelabs:
A little optimization for the `createLink` function on already existing links.
It will read the text from the a link instead of using the selection text from the document.

For me the current implementation sometimes results in
image

The new implementation gets rid of it


Alex-D:
Thanks :)

PR #726, opened 2018-02-27
[Feature] OpenModal - optional relative positioning

forelabs:
This feature enhancement will allow to open the modal relative to the current selected node,
which is quite useful when you have a lot of data in the editor and also use the autoGrow function so your content is far away from the top button pane.

In this case, when enabled and your selected node is part of your current viewPort, the modal opens relative to the current selected node (e.g. double clicked image).

image


Alex-D:
That's a lot of code for... not much :/

I want too keep Trumbowyg as light as possible, so I can't include that since it's not necessary and too weight to be negligible.

forelabs:
Okay :+1:

PR #727, opened 2018-03-01
Add options for minimal links and auto-prefixing http protocol

SDRACK:
Two options added: `autoPrefixHttpProtocol` and `minimalLinks`.

`minimalLinks` slims down the link overlay to use only `text` and `url`.

`autoPrefixHttpProtocol` adjust links without a protocol. Those with a protocol, anchors and relative links are left alone. Anything else if prepended with `http://` or `mailto:` as appropriate. 

Both new options default to `false`.

N.B. ES6 methods used for string interpolation. Can adjust if required.


src/trumbowyg.js
```diff
@@ -1232,6 +1248,15 @@ Object.defineProperty(jQuery.trumbowyg, 'defaultOptions', {
...
             });
         },
+        prependHttpProtocol: function(url) {
+            var t = this;
+            if(!t.o.autoPrefixHttpProtocol) { return url; }
+
+            const VALID_LINK_PREFIX = /^([a-z][-+.a-z0-9]*:|\/|#)/i;
+            if(VALID_LINK_PREFIX.test(url)) { return url; }
+
+            return url.includes("@") ? `mailto:${url}` : `http://${url}`;
```
Alex-D: That's wrong, you can use @ in HTTP url to put some user/password:

`[//[user[:password]@]host[:port]][/path][?query][#fragment]`

Alex-D:
Do not use ES6 since I did not use babel and want to be ES5-compatible.

SDRACK:
Cheers for the feedback - I'll update the ES6 bits and implement a regex to distinguish email addresses / urls.

Be back with you soon 👍

forelabs:
I suggest to rename it to autoPrefixProtocol and have an urlProtocol option which defaults to https, but can be set to e.g. Http. Https will be the new Standart since chrome will also mark websites as untrusted soon when they use http.

SDRACK:
Thanks, sounds sensible. I've updated the option name. Under the new setup, users can provide the following options to `urlProtocol`:

- `false` (default)
- `true`, `'https'` or `'https://'` => `'https://'`
- `'http'` or `'http://'` => `'http://'`

Anything else will be ignored.

Let me know what you think - appreciate your work on this!

src/trumbowyg.js
```diff
@@ -472,6 +474,23 @@ Object.defineProperty(jQuery.trumbowyg, 'defaultOptions', {
...
 
+        setupUrlPrefix: function() {
+          var t = this,
+              protocol = t.o.urlProtocol;
+
+          if(!protocol) { return; }
+
+          if(typeof(protocol) === 'boolean') {
+              return 'https://'
+          } else if(typeof(protocol) === 'string') {
```
Alex-D: Don't need else as we always manage all others cases with the previous if

src/trumbowyg.js
```diff
@@ -472,6 +474,23 @@ Object.defineProperty(jQuery.trumbowyg, 'defaultOptions', {
             this.btnsDef[btnName] = btnDef;
         },
 
+        setupUrlPrefix: function() {
+          var t = this,
+              protocol = t.o.urlProtocol;
+
+          if(!protocol) { return; }
+
+          if(typeof(protocol) === 'boolean') {
```
Alex-D: `if (typeof protocol !== 'string') {`

src/trumbowyg.js
```diff
@@ -1199,31 +1222,43 @@ Object.defineProperty(jQuery.trumbowyg, 'defaultOptions', {
...
-                    value: title
-                },
                 text: {
                     label: t.lang.text,
-                    value: text
-                },
-                target: {
-                    label: t.lang.target,
-                    value: target
+                    value: new XMLSerializer().serializeToString(documentSelection.getRangeAt(0).cloneContents())
```
forelabs: this one is out of date and should stay `text`

src/trumbowyg.js
```diff
@@ -480,16 +480,10 @@ Object.defineProperty(jQuery.trumbowyg, 'defaultOptions', {
...
-              if(/^https?:\/\/$/.test(protocol)) { return protocol; }
-
-              if(/^https?$/.test(protocol)) {
-                  return protocol + '://'
-              }
-          }
-        }
+          if(typeof(protocol) !== 'string') { return 'https://'; }
+          if(/^https?:\/\/$/.test(protocol)) { return protocol; }
+          if(/^https?$/.test(protocol)) { return protocol + '://'; }
```
Alex-D: That's a lot of code to just get `https://` or `http://` :/

Alex-D: What about others protocols? If I want put some ftp:// or whatever?

SDRACK: Thanks @Alex-D. 

Would you prefer to allow any string value in the option, and append `://` if it's missing?

So:

- a `true` value defaults to `https://`
- a string without the correct suffix has `://` appended to it
- a string with the correct suffix is prepended as is

Alex-D: yep I think it could be cool :)

SDRACK:
Morning @Alex-D - that's the `setupUrlPrefix` function shortened and set to allow any string as said prefix. For easy reference, here's the updated code:

```
        setupUrlPrefix: function() {
            var protocol = this.o.urlProtocol;
            if(!protocol) { return; }

            if(typeof(protocol) !== 'string') { return 'https://'; }
            return /:\/\/$/.test(protocol) ? protocol : protocol + '://';
        }
```

How does that look to you?

Alex-D:
LGTM :)

Could you update the doc about this new option? It's somewhere here: 

- https://github.com/Alex-D/Trumbowyg/blob/develop/docs/documentation/index.html#L64
- https://github.com/Alex-D/Trumbowyg/blob/develop/docs/documentation/index.html#L463

Thanks :)

SDRACK:
Fantastic - thanks @Alex-D. Really appreciate all your work with me on this. 

Let me know what you think of the updated docs and I'll edit if needed.

One thing to note: I've put `2.9.4` as the version added as I'm not sure of your process for iterating this. Let me know if that needs tweaking.

SDRACK:
@Alex-D - please don't consider this a chase, but I'm wondering what your process is for introducing new features such as these, and when you're next likely to do so?

Hope I'm not being a pain asking this :)

Alex-D:
Just taking time to run this on my computer, then merge other MRs, release a new version, test it, etc.

So I need to take couple of hours ^^'

Sorry if it's long...

SDRACK:
Thanks, that clears things up perfectly. It was more a case of establishing whether this would be deployed shortly or as part of a larger release way down the line. So that answers my question perfectly.

Thanks again for your work on this.

Alex-D:
Thanks! :)

jorrit:
This PR (and release 2.10.0) broke ES5 browsers/minifiers because of the use of `const`.

See #764.

PR #729, opened 2018-03-03
Added custom font size option

blufish:
History is not working here. Would not work using the trombowyg exexCmd function, so used jQuery instead.

### DIFF
--- a/plugins/fontsize/trumbowyg.fontsize.js
+++ b/plugins/fontsize/trumbowyg.fontsize.js
@@ -86,24 +86,18 @@
                                 value: '48px'
                             }
                         },
-                        // callback
                         function (values) {
-                            console.log("Got value back", values);
-                            //trumbowyg.execCmd('fontSize', values.size, true);
-                            
                             var text = trumbowyg.range.startContainer.parentElement
                             var selectedText = trumbowyg.getRangeText();
                             if($(text).html() == selectedText) {
                                 $(text).css('font-size', values.size);
                             } else {
-                                console.log("Creating new span for selected text");
                                 trumbowyg.range.deleteContents();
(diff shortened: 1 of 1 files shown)


Alex-D:
Remove commented code and console.log

blufish:
Updated

plugins/fontsize/trumbowyg.fontsize.js
```diff
@@ -75,6 +75,38 @@
...
+                            trumbowyg.saveRange();
+                            return true;
+                        }
+                    );
+                },
+                text: '<span style="font-size: medium;">Custom</span>',
+                hasIcon: false
+            };
+        trumbowyg.addBtnDef(freeSizeButtonName, freeSizeBtnDef);
+        dropdown.push('fontsize_custom');
```
pavelloz: Since its already assigned, i would use `freeSizeButtonName`:

```js
dropdown.push(freeSizeButtonName);
```

plugins/fontsize/trumbowyg.fontsize.js
```diff
@@ -75,6 +75,38 @@
...
+                fn: function () {
+                    trumbowyg.openModalInsert('Custom Font Size',
+                        {
+                            size: {
+                                label: 'Font Size',
+                                value: '48px'
+                            }
+                        },
+                        function (values) {
+                            var text = trumbowyg.range.startContainer.parentElement
```
pavelloz: I think you missed `;`

plugins/fontsize/trumbowyg.fontsize.js
blufish: I'll give this a shot

blufish:
This works, I added translations too.

Alex-D:
Thanks! :)

PR #737, opened 2018-03-12
[Plugin] Improve cleanpaste by only analyzing pasted content

forelabs:
I do not like that the cursor is resetted everytime you paste something,
so I have improved the cleanpaste plugin that it will:

- only analyze new pasted content
- add content right behind current cursor
- set cursor right after pasted content

### DIFF
--- a/plugins/cleanpaste/trumbowyg.cleanpaste.js
+++ b/plugins/cleanpaste/trumbowyg.cleanpaste.js
@@ -1,10 +1,11 @@
 /* ===========================================================
- * trumbowyg.cleanpaste.js v1.0
+ * trumbowyg.cleanpaste.js v1.1
  * Font Clean paste plugin for Trumbowyg
  * http://alex-d.github.com/Trumbowyg
  * ===========================================================
  * Authors : Eric Radin
  *           Todd Graham (slackwalker)
+ *           Sven Dunemann [dunemann@forelabs.eu]
  *
  * This plugin will perform a "cleaning" on any paste, in particular
  * it will clean pasted content of microsoft word document tags and classes.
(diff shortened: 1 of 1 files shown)


forelabs:
Btw  I use the insert part in a few plugins i wrote / modified.
Maybe it makes sense to extract it to a Trumbowyg core function?

Alex-D:
> Maybe it makes sense to extract it to a Trumbowyg core function?

Depends if it's usefull for the core. If not, no. If yes, we need to check that.

forelabs:
I think it is useful, there were already several requests / bug reports because of that: https://github.com/Alex-D/Trumbowyg/issues?utf8=✓&q=is%3Aissue+cursor+position

It's the wanted behaviour that when you insert something via plugin, it should be inserted where you at + you should stay with your cursor where you at. Getting the cursor set to start of editor is not what the user expects.

I will use single quotes and test it with browserstack on the mentioned browsers + on mobile devices too

Alex-D:
If it's light, I'm OK to add it to core, if core can benefit from it, it could be a good point! :)

Alex-D:
What's up there? :)

forelabs:
Sry had a lot of different stuff to do, I also have to test this one again, last time I used it I had some troubles with word content, but this plugin is exactly for that case so it should work. Sry to let you wait, will have a look into it the next days

dfreerksen:
What is the status of this?

This resolves the issue I am having where pasting content at the bottom of the content area moves the cursor and pastes it at the top instead.

Alex-D:
Need to be rebased anyway :/

Alex-D:
Maybe this one can be fixed too #805 ?

SadegOussama:
Hello,

I still have the same problem as described above(when I paste a text the cursor goes up, and when it is a link, the link is stuck at the top with the cursor), what can I do to solve it ? I imported cleanpaste well.

Thank you.

danreg:
any updates on this?

malandles:
Confirms that this PR works well on Chrome Version 75.0.3770.100

PR #745, opened 2018-03-22
[Plugin] Fix table plugin not working for multiple instances

chiiya:
Previously, as soon as you had more than one trumbowyg instance on the same page, the dropdown for the table plugin would only get added once.


Alex-D:
You right! Thanks! :)

PR #749, opened 2018-03-28
Add code highlight plugin by prismjs.

kirillurgant:
Add code highlight plugin by prismjs.
It depend of global Prism and prism.css. I did not include it in the repository, what would you advise in this case?


Globulopolis:
> I did not include it in the repository, what would you advise in this case?

Provide an option for this.

And it's better to do not use an ID for inputs to avoid possible conflicts if user page allready have tag with this ID value.

kirillurgant:
> Provide an option for this.
I can write instructions for this plugin.

kirillurgant:
I wrote the documentation. What do you think about this?

Alex-D:
Thanks! LGTM :)

PR #758, opened 2018-04-10
Fix urlProtocol option not taking effect

jdelStrother:
Without this fix, we generate a prefixed url with `prependUrlPrefix`, but don't use it in the newly-added link


Alex-D:
Good catch!

Thanks :)

Alex-D:
I will do that next week :)

PR #763, opened 2018-04-19
Fix broken insertHTML command in IE

### LINKED ISSUE #761: noembed not working in IE11
### Informations

**Browser version**: IE11 (and lower)
**OS**: Windows 7
**Screen resolution**: N/A

### How to reproduce the bug?
Visit the noembed demo in Edge. Use the button on the WYSIWYG to embed a YouTube video. When you click Confirm, the video is not embedded.

I'll look into this myself and submit a PR soon.

satchamo:
Fixes #761. It should also fix all the other plugins that use insertHTML.

### DIFF
--- a/src/trumbowyg.js
+++ b/src/trumbowyg.js
@@ -1381,7 +1381,14 @@ Object.defineProperty(jQuery.trumbowyg, 'defaultOptions', {
                         param = '<' + param + '>';
                     }
 
-                    t.doc.execCommand(cmd, false, param);
+                    var didWork = t.doc.execCommand(cmd, false, param);
+                    if(!didWork && cmd.toLowerCase() == "inserthtml"){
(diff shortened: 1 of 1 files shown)


Alex-D:
I think we can say that IE is dead now since Windows 7 is not supported anymore by Microsoft.

prjm:
I just wanted to note that Microsoft is still supporting IE on Windows 10, as listed [here](https://docs.microsoft.com/en-us/lifecycle/faq/internet-explorer-microsoft-edge). It is not recommended, but supported.

> Microsoft recommends using Microsoft Edge as your default browser and supports Internet Explorer 11 for backward compatibility.

I stumbled over this issue while using an embedded browser control and trying to call trumbowyg functions via the DOM-API of that embedded IE browser.

PR #765, opened 2018-04-19
workaround for Firefox incorrect selection of links

geonanorch:
Currently (Trumbowyg v2.10.0, Firefox 57 & 59.0.2) if one selects a link within the Trumbowyg editor (full link, e.g. by double-clicking on the link) and then activates the 'create link' dialog, link details are not populated (and the selection will be inaccurate after the dialog is closed).

This is due to a misbehavior in firefox, which in that particular case (full selection of 'A' element) returns the link's parent instead of the link's text.

This pull request addresses that particular case.

### DIFF
--- a/src/trumbowyg.js
+++ b/src/trumbowyg.js
@@ -1197,7 +1197,10 @@ Object.defineProperty(jQuery.trumbowyg, 'defaultOptions', {
                 title,
                 target;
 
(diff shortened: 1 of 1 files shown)


Alex-D:
This code has been rewritten since this PR.

I hope the bug is fixed with the rewrite.

Thank you for your PR :)

PR #766, opened 2018-04-21
Adding a fullscreen API option

### LINKED ISSUE #759: API feature: Trigger fullscreen mode from outside
### Informations
On mobile, I want to control the fullscreen mode, when the keyboard is up and down.
I was thinking I can do something like `trumbowygEl.trumbowyg({fullscreen: true})` but to no avail.

Is there any chance you add this functionality?

**OS**: Android, iOS, Chrome
**Screen resolution**: 425x425

Stanwar:
Adding a fullscreen option to the API. This can be used as 
editor.trumbowyg('fullscreen', true).

Fixes #759

### DIFF
--- a/src/trumbowyg.js
+++ b/src/trumbowyg.js
@@ -174,6 +174,10 @@ Object.defineProperty(jQuery.trumbowyg, 'defaultOptions', {
                     case 'empty':
                         return t.empty();
 
+                    // Full Screen
+                    case 'fullscreen':
+                        return t.fullscreenMode(params);
+
                     // HTML
                     case 'html':
                         return t.html(params);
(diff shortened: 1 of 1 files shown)


src/trumbowyg.js
```diff
@@ -1054,6 +1058,14 @@ Object.defineProperty(jQuery.trumbowyg, 'defaultOptions', {
             }
         },
 
+        // Fullscreen Mode management
+        fullscreenMode: function(isFullScreen){
+            var t = this;
+            // Check if fullscreen mode.
+            if(isFullScreen){
+                t.fullscreen();
```
Alex-D: This can't work as is :/

fullscreen method toggle the fullscreen state, so the actual behaviour is weird.

Stanwar: @Alex-D Could you elaborate on the weird behavior? I tried this in a project of mine with fullscreen option enabled by default with a fullscreen button as well. Worked for me. If you could elaborate, I dont mind going back and making changes.

Stanwar: @Alex-D Thanks for the insight. Ill go back to the drawing board.

wilmaribg:
use 

el.trumbowyg('execCmd', {cmd: 'fullscreen'})

src/trumbowyg.js
wilmaribg: is same and functional
 
el.trumbowyg('execCmd', {cmd: 'fullscreen'})

Issue #775, opened 2018-05-01
When editor is inside a form, buttons don't appear in Insert Link modal


ppritcha:
Trumbowyg is a great lightweight editor, and meets my needs perfectly except for one serious problem.  When the editor is used inside a form, everything works except for the "Insert link" dialog.  The dialog appears, but the "Confirm" and "Cancel" buttons are not displayed, so the modal cannot be closed properly.

The problem appears to be that the modal dialog is a form appended to the editor container.  Because the editor is already inside a form, this results in a form inside another form (or a "nested form"), which is not valid HTML.  The fix for this problem would be to append the modal dialog to the document rather than to to editor container.

Note that this is related to issue #20 and issue #126.

Alex-D:
As you can read in #126:

> I don't want to fix that, It need lot of works, sorry.

I will fix that in a v3, later, but I did not have the time to work on that for now.

PR #779, opened 2018-05-02
Append modal to document body

ppritcha:
Fix for #775, "When editor is inside a form, buttons don't appear in Insert Link modal".

### DIFF
--- a/src/trumbowyg.js
+++ b/src/trumbowyg.js
@@ -1418,8 +1418,8 @@ Object.defineProperty(jQuery.trumbowyg, 'defaultOptions', {
             var $modal = $('<div/>', {
                 class: prefix + 'modal ' + prefix + 'fixed-top'
             }).css({
-                top: t.$btnPane.height()
-            }).appendTo(t.$box);
+                top: t.$box.height() + t.$btnPane.height()
+            }).appendTo($(document.body));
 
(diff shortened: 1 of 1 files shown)


Alex-D:
Did you test fullscreen mode?

src/trumbowyg.js
```diff
@@ -1418,7 +1418,8 @@ Object.defineProperty(jQuery.trumbowyg, 'defaultOptions', {
             var $modal = $('<div/>', {
                 class: prefix + 'modal ' + prefix + 'fixed-top'
             }).css({
-                top: t.$box.height() + t.$btnPane.height()
+                top: t.$box.offset().top + t.$btnPane.height(),
+                zIndex: 99999
```
Alex-D: Why so high value?

I'm pretty sure we can set a reasonable value here.

ppritcha: It's set to 99999 so that it will appear on top of the fullscreen editor, which you have set to 99999 in your CSS.

    .trumbowyg-fullscreen {
      position: fixed;
      top: 0;
      left: 0;
      width: 100%;
      height: 100%;
      margin: 0;
      padding: 0;
      z-index: 99999; }

Alex-D: Oh, Okay. My bad, sorry.

PR #782, opened 2018-05-10
Added support for maxlength

suilillo:
If the textarea has maxlength attribute then it will applied to editor


Alex-D:
It's a lot of code to just manage a limit :/

PR #784, opened 2018-05-16
Fix ViewHTML button not working in disabled state

### LINKED ISSUE #780: View HTML button do nothing when editor is disabled
### Informations
if editor is disabled - either by specifying html attribute or passing as option, the **View HTML** button is still enabled, but it do nothing when you click the button.
Should **View HTML** button also be disabled, or it should able to toggle between HTML and WYSIWYG mode?

**Browser version**: Chrome 66
**OS**: Mac
**Screen resolution**: 1440 x 900

### How to reproduce the bug? 
1. Visit https://codepen.io/anon/pen/yjXEoB?editors=1010
2. Click **View HTML** button

ethan-ho:
As titled.

Fixes #780

### DIFF
--- a/src/trumbowyg.js
+++ b/src/trumbowyg.js
@@ -268,7 +268,8 @@ Object.defineProperty(jQuery.trumbowyg, 'defaultOptions', {
             };
         t.btnsDef = {
             viewHTML: {
-                fn: 'toggle'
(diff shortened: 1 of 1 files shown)


Alex-D:
Thanks :)

PR #807, opened 2018-06-28
Danish translations

KTW-NIRAS:
Added and modified some Danish translations.
Should be 100% translated now.


Alex-D:
LGTM Thanks :)

PR #825, opened 2018-08-07
Use HTML5 Color Picker

BourotBenjamin:
Hi,

As wysiwyg editors are to make it easier to post formated content to everyone, we couldn't expect from them to know / understand colors hex codes.

So it would be much easier for the to use a color picker.


Alex-D:
Does it works in any browsers? Does it fallback to text if not supported?

Alex-D:
OK, it works in a lot of browsers and fallbacks to input type text if not.

Thanks for the improvement! :)

PR #827, opened 2018-08-11
Use native emoji code points instead

zanechua:
Was trying to implement the emoji plugin, ran into some issues (emojione/emojify.js#175) and figured implementing native emoji's would probably be better than having to:

1. Have javascript run to replace the emojis every single time
2. Use an external library
3. Use an unmaintained external library
4. Load external image files for every single emoji

Emoji List is taken from: https://github.com/emojione/emojione/blob/master/emoji.json

I wrote a small nodejs script to retrieve all the emoji code points sans the skin color versions as not all OSes have support for that. Would you like this script? Should probably be added into the build process somewhere to automate it.


Alex-D:
> Would you like this script? Should probably be added into the build process somewhere to automate it.

Yep :)

Can you update the related documentation and demo?

Here:
https://github.com/Alex-D/Trumbowyg/blob/develop/docs/demos/plugins/emoji.html

And here:
https://github.com/Alex-D/Trumbowyg/blob/develop/docs/documentation/plugins/index.html#L219-L266

zanechua:
Done.

zanechua:
The script to download the file and output to html codepoints is below.

I added the current set of html code points that are being used here.

https://gist.github.com/zanechua/b22559644391f76f276f0a3d4d6e946b

dmyers:
Native emojis would be amazing! I also had a hard time getting the emoji plugin to work with emojify.js and finally did by copying all the files from the demo since somehow getting the latest versions from all libraries was not working at all. Once I got it setup, I just am not as fond of it since users can't copy between apps like messages into the editor and have it work since this implementation is png based instead of utf codes.

dmyers:
Have you by chance seen Emoji Mart? It looks pretty neat and supports native too.

https://missive.github.io/emoji-mart/
https://github.com/missive/emoji-mart/

dmyers:
I just tried it now. It works good and I like it too. I think that's a good solution for me at least.

PR #831, opened 2018-08-15
Fix fontsize modal localization and add more translation for zh_tw

HouseAlwaysWin:
I found I can't localize custom font size modal and some of plugins don't have a translation for traditional chinese...


plugins/fontsize/trumbowyg.fontsize.js
```diff
@@ -1,4 +1,4 @@
-(function ($) {
```
Alex-D: I'm not sure about that :-°

plugins/fontsize/trumbowyg.fontsize.js
```diff
@@ -105,11 +110,10 @@
         var freeSizeButtonName = 'fontsize_custom',
             freeSizeBtnDef = {
                 fn: function () {
-                    trumbowyg.openModalInsert('Custom Font Size',
-                        {
```
Alex-D: Keep this on a new line.

src/langs/zh_tw.js
```diff
@@ -10,48 +10,63 @@
...
+    lineheight: '間距大小',
+    lineheights: {
+        '0.9': '間距小',
+        'normal': '預設間距',
+        '1.5': '間距中',
+        '2.0': '間距大'
+    },
+    fontFamily: '字體',
+    fontsize: '字體大小',
+    fontsizes: {
```
Alex-D: Do not edit this file.

Translations for plugins should be **IN** the plugin file since they are useless for the core.

src/langs/zh_tw.js
```diff
@@ -10,63 +10,39 @@
...
+    fullscreen: '全螢幕',
+    close: '關閉',
+    submit: '確定',
+    reset: '取消',
+    required: '必需的',
+    description: '描述',
+    title: '標題',
+    text: '文字',
+    superscript:'上標',
+    subscript:'下標',
```
HouseAlwaysWin: I found original zh_tw.js doesn't have superscript and subscript translations but also other language doesn't have either, and I didn't see the plugin for superscript and subscript so I added it to this file

src/trumbowyg.js
```diff
@@ -1382,6 +1382,11 @@ Object.defineProperty(jQuery.trumbowyg, 'defaultOptions', {
                         param = '<' + param + '>';
                     }
 
+                    var del = t.$ed.find('del')[0];
+
+                    if (del) {
+                        t.$ed.find('del').contents().unwrap();
+                    }
```
HouseAlwaysWin: I think this is not quite good solution but I can't find why execCommand doesn't work for the del button toggle on so I add this exception for at least can work.

HouseAlwaysWin:
done, remove commits:

- Fix bug del button can't toggle on

- fix strikethrough bug

PR #832, opened 2018-08-22
Fix strikethrough toggle button bug

HouseAlwaysWin:
strikethrough button can't remove strike after activate


Alex-D:
You should just switch `semantic` to false.

Your fix is not a fix, sorry.

PR #833, opened 2018-08-23
fix when semantic config is false, the button panel is not work correctly

HouseAlwaysWin:
I found when I set semantic to false the button panel will not work correctly,
screenshot_20180823_153442

and in semantic false mode IE will use strong and em instead of b and i tag


Alex-D:
Short: I don't know.
Long: I need to take the time to check that and why differents tags are used, and why Trumbowyg does not normalize them.

Sorry but for now I can't work a lot on Trumbowyg, but I will have a lot of time soon :)

PR #846, opened 2018-09-17
add fontFamily name in button pane

SerhiyP:
add fontFamily name in button pane when click or select text in the editor and it has font style attribute
I change dependencies because I can't start project


Alex-D:
I think the default value should keep the "Font family" translated label.

Since it's a WYSIWYG, by defaut, that's not a specific font, but a CSS defined font which can change.

SerhiyP:
> I think the default value should keep the "Font family" translated label.
> 
> Since it's a WYSIWYG, by defaut, that's not a specific font, but a CSS defined font which can change.

Yes, I think about it. But I use another approach to solve this, I create the function that takes a fontFamily from BODY

But now I think from another way do this

maybe I should use this method **updateButtonPaneStatus**
What do you think about it&

Alex-D:
Needs some rebase :/

Issue #703, opened 2018-01-27
When all tags are empty, Trumbowyg cleans content but should keep it


noistudio:
### Informations
Try to insert .
`   <a href="#"><i class="fa fa-facebook"></i></a>
                                <a href="#"><i class="fa fa-twitter"></i></a>
                                <a href="#"><i class="fa fa-instagram"></i></a>
                                <a href="#"><i class="fa fa-behance"></i></a>
                                <a href="#"><i class="fa fa-vk"></i></a>
                                <a href="#"><i class="fa fa-youtube"></i></a>`

After press show html.
After that return back,and you can see the editor is empty

Alex-D:
Yep, because there is not text. Si Trumbowyg thing all that empty tags are not wanted and empty the editor.

forelabs:
Maybe add an option like `keepEmptyTags` would be a possible solution? Because I think in the most cases the current behaviour is the wanted behaviour.

Your inserted html is part of a syntax which works for you environment, because you have font-awesome installed.

I think for a normal user it makes more sense to use images and link those images instead rely on css elements which are only present on the current page.

dfreerksen:
Where in the code are empty tags being stripped out? I need this too so I would be happy to add the `keepEmptyTags` option.

qJake:
Just wanted to say thanks for adding this - our customers often use this to add simple containers like `<div class="my-custom-container"></div>` and then style it with some custom CSS.

Much appreciated!

Alex-D:
Check the 2.12.0 release notes.

New option here: https://alex-d.github.io/Trumbowyg/documentation/#tags-to-keep

PR #851, opened 2018-09-26
Introduce keepEmptyTags option

dfreerksen:
This came from issue #703 where if the content contains only empty tags, the content gets completely removed. It currently assumes if there is no text, the tags are not wanted or necessary. This is not always the case.

This introduces a new option for `keepEmptyTags`. To stay consistent with current functionality, the default value is `false`. Setting the value to `true`, the content will not be reset to an empty string even when there is not text in the tags.


Alex-D:
What about #861

PR #861, opened 2018-10-12
Bugfix for empty $oldTag

kiewietr:
empty oldTag tags get surrounded by newTag on every keystroke. This resulted in huge content. I dont know what the best solution would be but I added a simple contents().length check and only replace it when there is content inside the tag.

This would occur with <i> tags when using font-awesome for example.

### DIFF
--- a/src/trumbowyg.js
+++ b/src/trumbowyg.js
@@ -1178,6 +1178,7 @@ Object.defineProperty(jQuery.trumbowyg, 'defaultOptions', {
 
             $(oldTag, this.$ed).each(function () {
                 var $oldTag = $(this);
+                if($oldTag.contents().length == 0) return false
                 $oldTag.wrap('<' + newTag + '/>');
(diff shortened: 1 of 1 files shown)


Alex-D:
Hum, what about: #851 
What's the best solution?

The original idea is to delete all empty tags, because there are useless.

kiewietr:
> The original idea is to delete all empty tags, because there are useless.

Sorry didnt see that one. I think it's an okay solution if you also have the keepEmptyTags option or some option to exclude certain tags from deletion, because we use font-awesome and these icons do use empty i tags.

`<i class="fa fa-check"></i>`

Alex-D:
I understand, there is a list of tags to not delete here: https://github.com/Alex-D/Trumbowyg/blob/v2.11.1/src/trumbowyg.js#L1074

Alex-D:
Ok, I add this to a fix I do.

Thanks.

PR #867, opened 2018-10-18
Update the loader for the demos

zanechua:
Update the loader so that the demo for emojis are working + changed from rawgit.com as it has been deprecated.


Alex-D:
Thanks! Did not see rawgit deprecation message before.

PR #879, opened 2018-11-08
Allowing script tags with src to be editable.

nurogenic:


Alex-D:
Fixed in another way in v2.12. See https://alex-d.github.io/Trumbowyg/documentation/#tags-to-keep for more informations.

Issue #881, opened 2018-11-16
Cleanpaste cleans the already inserted text, not the newly pasted text.


Geniekort:
When trying to use the Cleanpaste plugin (which is awesome by the way), the wrong text is cleaned; 

So when I paste some text which is not clean (for example it has a `style` attribute), the style attribute is pasted anyway, so the cleanpaste plugin is not picking it up.

However when I paste something again, next to the already pasted stuff, then the first text is cleaned.

To reproduce:

1.  I copy `<p style="color:red;">This goes wrong</p>`from a rendered html page. (So you actually see red text)

2.  I paste it in an initialized Trumbowyg editor (Observe it is still red)

3.  I copy `<p>Bladibla</p>` from a rendered html page

4.  I paste it in the same Trumbowyg editor. 

5. Now the "This goes wrong" text gets cleaned.

I can reproduce this on Ubuntu, in Chrome, Firefox and Opera.

Any suggestions?

Geniekort:
Thanks! I"ll check if it's working now!

Geniekort:
@Alex-D isn't the current behaviour "wrong"? I mean, cleaning all the text inside the text box seems fine, but the pasted text itself should also be cleaned right?

Alex-D:
Deployed in 2.12.1

PR #883, opened 2018-11-19
Fix for cleanpaste

Geniekort:
Solving the bug as described in #881 , by adding a timeout to the execution in of the cleanpaste plugin.

### DIFF
--- a/plugins/cleanpaste/trumbowyg.cleanpaste.js
+++ b/plugins/cleanpaste/trumbowyg.cleanpaste.js
@@ -97,14 +97,14 @@
         plugins: {
             cleanPaste: {
                 init: function (trumbowyg) {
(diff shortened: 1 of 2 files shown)


Alex-D:
What are you trying to fix like that? You are just delaying a push in an array :/

That's the moment when the plugin is registered into Trumbowyg. Are you sure to load `trumbowyg` *then* `trumbowyg.cleanpaste` ?

Geniekort:
You are right, this should fix it

Alex-D:
Did you test it?

Geniekort:
Yes I tested it. It would be nice if we had some specs to run for Trumbowyg. Nice to see that you are actively developing it!

Alex-D:
I hope unit-test/spec for v3 :)

Issue #885, opened 2018-11-24
Change event not fired when image is moved around the editor


dolbex:
### Informations

**Browser version**: Chrome 70
**OS**: Mac OS
**Screen resolution**: 1600 

### How to reproduce the bug?

If you drag an image around the editor there is no update fired. Checking the source fires the update so I can't tell if the source is updated. I would imagine it is since the image appears in the place I would suspect. 

I know you are working on the next version but a resolution on this issue would solve one of our final issues on our own open source project that relies on Trumbowyg: The Greatest Editor of All Times. :)

Alex-D:
Arf, I've never test this :/

Maybe it could fix by adding "drop" listener here: https://github.com/Alex-D/Trumbowyg/blob/v2.12.0/src/trumbowyg.js#L690

dolbex:
Cool - thanks for the direction. Posted a pull for this https://github.com/Alex-D/Trumbowyg/pull/886

Alex-D:
Fixed in 2.12.2 :)

PR #886, opened 2018-11-25
Added drop listener to announce changes when moving images in editor

dolbex:
Addressing issue here: [#885](https://github.com/Alex-D/Trumbowyg/issues/885) so that when a user drags an image around the editor it fires the appropriate change event.


dist/trumbowyg.js
```diff
@@ -653,6 +653,12 @@ Object.defineProperty(jQuery.trumbowyg, 'defaultOptions', {
                         }
                     }
                 })
```
Alex-D: You should merge drop and cut listener since they have the same behavior:

```
.on('cut drop', ...
```

Alex-D:
I've done the fix! :)

dist/trumbowyg.js
dolbex: Will do sir.

dolbex:
Thanks so much Alex.

PR #887, opened 2018-11-26
Improved Dutch translation

avaneerd:
Added missing translations and changed the translation for 'reset' to more closely reflect it's usage.


Geniekort:
Seems like a better translation!

Alex-D:
Thanks :)

Issue #417, opened 2016-10-10
Ctrl+Z does not work without the Undo button


ilpoli:
Chrome 53, Trumbowyg 2.3.0
1. Configure Trumbowyg without Undo button.
2. Type some text
3. Use Ctrl+Z to undo last changes

**Expected**: Textarea values is updated.
**Actual**: Textarea still contains value as before Ctrl+Z.

As workaround we added the Undo button and hided it.

Alex-D:
Undo feature is not stable and the button only works on Blink/V8 browsers.

ilpoli:
Thanks for explanation. Are you planning to improve this logic closer time or it is not a priority feature?

Alex-D:
That's not a priority.

Alex-D:
Since 2.10.0, you can use history plugin to get an improved and cross-browser history management.

https://alex-d.github.io/Trumbowyg/documentation/plugins/#plugin-history

Issue #227, opened 2016-02-10
Cut, undo, redo is not working properly


saggiyogesh:
After Cut, undo, redo, `tbwchange` event is not giving correct html.

Alex-D:
Curious. I need to check that. Which version do you use? Last beta release or develop branch?

saggiyogesh:
latest. as i see code there is no handler for cut event.

Alex-D:
Nope. It's managed by the browser :)

Alex-D:
I've found a workaroud which is to listen Ctrl+Z / Ctrl+Y keyup events and `cut` event.

It's only on `develop` branch for now. I hope it was what you expected.

PR #888, opened 2018-11-27
Fix issue with undo/redo don't update textarea.

sgolomedov:
Force to update textarea content after undo/redo.
Issues #227, #417.


Alex-D:
Thanks!

PR #907, opened 2018-12-27
Move dropdown div under button-group div

wenpress:
Reads better semantically than adding after textarea.
Provides support for using position:sticky with top:some_offset on buttonbar
(dropdown appears in correct location when doing so)

### DIFF
--- a/src/trumbowyg.js
+++ b/src/trumbowyg.js
@@ -1426,7 +1426,7 @@ Object.defineProperty(jQuery.trumbowyg, 'defaultOptions', {
             var $modal = $('<div/>', {
                 class: prefix + 'modal ' + prefix + 'fixed-top'
             }).css({
-                top: t.$box.offset().top + t.$btnPane.height(),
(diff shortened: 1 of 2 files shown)


src/trumbowyg.js
```diff
@@ -805,7 +805,7 @@ Object.defineProperty(jQuery.trumbowyg, 'defaultOptions', {
                         $dropdown.append(t.buildSubBtn(def));
                     }
                 });
-                t.$box.append($dropdown.hide());
+                $btn = $btn.add($dropdown.hide());
```
Alex-D: Why not just `$btn.append` here?

wenpress: My dated javascript knowledge...  :/
I can do a further commit to use append or you can double check nothing is broken under this concpot first?

Alex-D:
That's not semantic to put div into button. And some things are not working on my side. I prefer keep the code as is.

PR #909, opened 2018-12-28
Add h5, h6 & pre to formatting dropdown

wenpress:
Added h5 and h6 to the formatting dropdown as they were missing.
Also added pre to the same dropdown as it's a basic text block style and the preformatted plugin actually adds a pre wrapped code block which is different.
Added what translations I could for "Preformatted".


Alex-D:
Is that really usefull in core?

I think it's not used enough to be added in core. You can do the same thing by adding custom buttons and something like that, no?

Also, is the `pre` supported by any browser? (I did not check)

wenpress:
I'm assuming you're talking about the pre tag being a separate plugin vs the h5 and h6 tags as obviously they need to be grouped with the existing h1-4 in the core otherwise it's a bad user experience.

So as for pre, it's supported by all browsers, the official spec is here:
https://www.w3.org/TR/html50/grouping-content.html#the-pre-element

It's used by a lot of people (being a very old tag) and not just for code listings.

For example, if you wish to display preformatted output (eg the output of a command) or anything else with preserved tabs, line breaks and whitespace.  It's a paragraph level tag as per the others in the core formatting dropdown.

Some people use it for displaying blocks of code however we now have a code tag for that which is more semantic.  However there are several problems with the code tag as defined by the w3c...

If you wish to include a command in the middle of a sentence as part of instructions you can surround it with a code tag, and as the tag is an inline element all is well.
However, if you wish to write a block of code the convention is to include the code tags within pre tags to turn the text marked semantically as code into a whitespace protected block element. This is what the "preformatted" plugin does. Due to the name of that plugin I expected to get only pre tags, nopt code tags as well, so it broke the layout, hence I realised there was no current way to get just basic pre tags.

If you examine TinyMCE you can see pre is listed as a paragraph type, along with basic paragraph and header tags. As it is a blocked paragraph element I feel it belongs with the others in the core dropdown for best user experience and match the convention used elsewhere.

(however I would, and will, make a code tag a plugin).

Alex-D:
For `pre` I was talking about execCommand support, not browser support :p Anyway it seems OK :) So I'm OK with that.

For `h5` and `h6`, I'm OK to add them in definition, but I do not think it's usefull to add them in default button list.

wenpress:
I was a bit confused when pre and browser support, lol...  Sorry!

All makes sense including not making the h5 and h6 appear by default. If the definitions are there how do I specify them in the startup options for the Formatting menu to include them as they're not paert of the default?

Alex-D:
You just need to create your own button which is a dropdown with all buttons you want in it, then use it in button list you want :)

Alex-D:
Reworked and merged in https://github.com/Alex-D/Trumbowyg/commit/5b09f32a59b56df638cccb80c914de23c7355109

clwilliams8:
@Alex-D -- I'm trying to get the paragraph formatting dropdown to include:
 "H5 Heading 5"
"H6 Heading 6"

I see you said to create your own button, and I've read the documentation on doing that, but having a hard time. Is there anyway you could provide code on how to include these two options?

Thank you so much

Issue #196, opened 2015-11-02
IE error - Could not complete the operation due to error 800a025e 


rvwoens:
I got this error in IE when loading a trumbowyg element using ajax. 

It turned out it was due to a removeAllRanges without checking. This fixed it:

```javascript
if (s.rangeCount > 0 && s.getRangeAt(0).getClientRects().length > 0) {
    s.removeAllRanges();
}
```

Alex-D:
Hi!

Which version of IE?

mrtom808:
Hi, this problem is present in the latest version. Can you re-apply the fix? (IE11)

https://github.com/Alex-D/Trumbowyg/blob/develop/src/trumbowyg.js#L1513

PR #915, opened 2019-01-16
Fix IE11 Error Could not complete the operation.

sgolomedov:
Fix IE11 Error 'Could not complete the operation due to error 800a025e', that sometimes appears on start.
https://stackoverflow.com/questions/16160996/could-not-complete-the-operation-due-to-error-800a025e
Fix Issue https://github.com/Alex-D/Trumbowyg/issues/196


Alex-D:
Why not just catch any error with a try-catch block? It will be more safe I think.

M0ns1gn0r:
Unconditional `catch` without analyzing the specific error/exception is considered bad practice.

Alex-D:
Yep but it's small (less chars). Since Trumbowyg should keep everything small as possible, I think try-catch should be OK.

sgolomedov:
@Alex-D Short try/catch version added.

Alex-D:
Thanks :)

Issue #901, opened 2018-12-19
Image Resize Plugin breaks after History or Paste Action


realulim:
### Informations

Tested with newest Firefox and Chrome on MacOS.

### How to reproduce the bug?
1. Go to a Page with activated Resize Plugin (Live Demo page at https://alex-d.github.io/Trumbowyg/demos/#plugins-resizimg will do)
2. Delete Image (place cursor behind it and press Backspace)
3. Undo delete via History plugin (if you have History plugin installed)
4. If you don't have the History plugin installed, you can also simulate the effect with ctrl-x ctrl-v (cut and paste)
5. Image is restored after either undelete or paste, but it cannot be resized anymore

I could not see a relevant difference between the two DOMs (resizable and not resizable), so I suspect that some Javascript initialisation code would have to re-run to make the image resizable again. If true, then this would probably mean that images would also not be resizable after saving and loading the editor content.

realulim:
My last assumption turned out to be correct. Even when reloading the page with the editor, which should cause all the JavaScript to run again, the image - once unresizable - never becomes resizable again.
This discovery led me to a workaround: a non-resizable image has the class "resizable". Oddly enough, removing this class make the image resizable again, even though this class is immediately reassigned.

realulim:
Looking at the code (https://github.com/Alex-D/Trumbowyg/blob/develop/plugins/resizimg/trumbowyg.resizimg.js) it seems line 48 is to blame. Only images that do not already have the class "resizable" are initialised by calling the resizable function from the jquery-resizable plugin. So if you save a resized image and load it later, it won't be initialised, because it already has the CSS class. Workaround is obviously to strip the class before saving.
I'm not sure what the best way to fix this bug is, otherwise I'd submit a pull request. But I suppose that there should be a better way to check, whether an image has already been initialised other than checking for the existence of the CSS class "resizable".

Alex-D:
I assume it's done like that to avoid re-init the resizable plugin on each image each time.

And yes, remove all resizable class should be a good solution: maybe at plugin init. Also it should be renamed with trumbowyg prefix to avoid conflicts.

realulim:
The resizable class is supplied by the jquery-resizable plugin, so there's nothing we can do about the name. If we want to fix this bug, then it's not enough to just remove the class at plugin init, because as soon as an image is deleted and restored or cut and pasted, the resizable class will reappear.

[...]

PR #919, opened 2019-01-26
Further Fix for #901

realulim:
Upon destruction of the Resizable, we need to remove the class 'resizable', otherwise a subsequent init will not work.

### DIFF
--- a/plugins/resizimg/trumbowyg.resizimg.js
+++ b/plugins/resizimg/trumbowyg.resizimg.js
@@ -38,7 +38,7 @@
                                     return false;
                                 },
                                 onDragEnd: function () {
-                                    trumbowyg.$c.trigger('tbwchange');
+                                    trumbowyg.$c.trumbowyg('execCmd', {cmd: 'syncCode'});
                                 }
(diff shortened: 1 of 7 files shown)


plugins/resizimg/trumbowyg.resizimg.js
```diff
@@ -53,7 +53,8 @@
                     function destroyResizable() {
                         trumbowyg.$ed.find('img.resizable')
                             .resizable('destroy')
-                            .off('mousedown', preventDefault);
+                            .off('mousedown', preventDefault)
+			    .removeClass('resizable');
```
Alex-D: Please use spaces for indent

realulim:
Fixed it. Perhaps this would be a good time to do another release?

Alex-D:
Done with 2.14.0 https://github.com/Alex-D/Trumbowyg/releases/tag/v2.14.0

realulim:
Splendid, many thanks!

realulim:
Just need to remember you don't like tabs :)

PR #922, opened 2019-02-01
Keep attributes when converting to semantic elements

### LINKED ISSUE #921: Inline styles removed when converting to semantic elements
I found another small but slightly annoying bug (or feature), which is the fact that inline styles are removed when the editor is converting non semantic elements to semantic elements.

### Informations

**Browser version**: Chrome Version 71.0.3578.98 (Official Build) (64-bit)
**OS**: macOS High Sierra (10.13.6 (17G4015))
**Screen resolution**: 1440x900

### How to reproduce the bug?

- Open a default trumbowyg instance
- Open `viewHTML`
- Paste the following content `<b style="color: rgb(230, 100, 84); font-size: x-large; line-height: 34.2857px;">Some text that I want in red with x-large fontsize</b>`
- Close `viewHTML`
- The styling will be removed
- Try with the following content `<div style="color: rgb(230, 100, 84); font-size: x-large; line-height: 34.2857px;">Some text that I want in red with x-large fontsize</div>`
- Close `viewHTML`
- The styling will not be removed

### Gif animation of the behaviour

trumbowyg

### Reason for the behaviour

If semantic is set to true, which is by default, then `b`, `i`, `s` and `strike` tag will be converted to their semantic tags `strong`, `em` and `del`. Whilst doing so, the attributes set to these tags will not be passed to their new semantic tag, which means you lose the styling. See https://github.com/Alex-D/Trumbowyg/blob/develop/src/trumbowyg.js#L1119.

I think these attributes should be passed though, so you keep the styling whilst converting the elements to something with a semantic meaning. I took the liberty to open a PR, see https://github.com/Alex-D/Trumbowyg/pull/922.

Let me know what you think!

Joralf:
Closes #921

If semantic is set to true, which is by default, then `b`, `i`, `s` and `strike` tag will be converted to their semantic tags `strong`, `em` and `del`. Whilst doing so, the attributes set to these tags will not be passed to their new semantic tag, which means you lose the styling. See https://github.com/Alex-D/Trumbowyg/blob/develop/src/trumbowyg.js#L1119.

I think these attributes should be passed though, so you keep the styling whilst converting the elements to something with a semantic meaning. I took the liberty to open a PR: , let me know what you think!

### DIFF
--- a/src/trumbowyg.js
+++ b/src/trumbowyg.js
@@ -1117,10 +1117,10 @@ Object.defineProperty(jQuery.trumbowyg, 'defaultOptions', {
             t.syncCode(force);
 
             if (t.o.semantic) {
-                t.semanticTag('b');
-                t.semanticTag('i');
-                t.semanticTag('s');
-                t.semanticTag('strike');
+                t.semanticTag('b', true);
+                t.semanticTag('i', true);
+                t.semanticTag('s', true);
+                t.semanticTag('strike', true);
 
                 if (full) {
                     var inlineElementsSelector = t.o.inlineElementsSelector,
--- a/src/trumbowyg.js
+++ b/src/trumbowyg.js
@@ -79,6 +79,7 @@ Object.defineProperty(jQuery.trumbowyg, 'defaultOptions', {
         prefix: 'trumbowyg-',
 
         semantic: true,
+        semanticKeepAttributes: false,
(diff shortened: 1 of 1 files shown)


Joralf:
Hi @Alex-D ,

I updated the PR! 👍

PR #923, opened 2019-02-04
Add indent/outdent functionality using Tab/Shift+Tab

tyronx:
This works for me on Google chrome.


tyronx:
Possible, but don't you think this should be the default behavior? All other major WYSIWYG editors do it like that as well.

tyronx:
Option added and off by default.

Alex-D:
Thanks :)

Daenero:
@Alex-D 
Can you please build & publish latest version to npm?
Thanks!

Alex-D:
I will do that soon :)

Issue #638, opened 2017-10-08
Delete/Strike-through button partially working


jpdelmundo:
### Informations

**Browser version**: Google Chrome Version 61.0.3163.100 (Official Build) (64-bit)
**OS**: Windows 10 Pro
**Screen resolution**: 1920x1080

### How to reproduce the bug?
1. Highlight/select text from the editor
2. Click on the delete/strike-through button (text style changed to strike-through)
3. Highlight/select the same text from the editor
4. Click on the delete/strike-through button (strike-through style not removed)

Video in the attached zip
[delete-button-bug.zip](https://github.com/Alex-D/Trumbowyg/files/1366845/delete-button-bug.zip)

Alex-D:
I can't fix that :/

Sorry.

jpdelmundo:
Sorry for asking. Is it because it is not a valid bug? Or just have no time fixing?

If it's latter, this issue should at least stay open for others see and probably someone might be able to fix it.

Alex-D:
It's a browser-level bug, so I can't fix it easily :/

jpdelmundo:
I see. I don't want to sound pushy about this but, if you can't fix it easily, can you at least leave it open for others to see?

Alex-D:
Nope. Because if I do that with all bugs like this I get 100+ open issues which stay like that forever :/

CharlotteBeebee:
Unless you specifically need `<del></del>` tags then this can be fixed by using `<s></s>` tags: in semanticCode function, replace t.semanticTag('strike', 'del') with t.semanticTag('strike', 's')

tomsx:
@CharlotteBeebee: I agree. <s></s> works just fine.
e.g. https://www.froala.com/wysiwyg-editor
...but <del></del> works in other wysiwyg editors as well. https://trix-editor.org/

davidsneal:
Changing the value of these two lines to `'s'` does the trick for me: https://github.com/Alex-D/Trumbowyg/blob/develop/src/trumbowyg.js#L441-L442

tyronx:
Why was a fix for this never added? Why are issues just being closed without even looking at possible fixes? So many questions...

tomsx:
Strangely the issue only occurs when trumbowyg.min.js is used. It works all fine with  trumbowyg.js.

darinredding:
Not working for me using either js version using zip downloaded from master today.

PR #929, opened 2019-02-13
Make strikethrough formatting removable

tyronx:
Basically a fix for issue #638

- Fix Strikethrough button naming
- Under default settings, show the Strikethrough button instead of the Deleted button
- Don't use obsolete `<strike>` tag
- Don't use `<del>` tag, because it can't be removed using removeFormat

Tested and working on google chrome. Backwards compatible. (users using the del plugin still can use that one )

Fix only works when users manually replace their pre-existing `<del>` tags with `<s>` tags and update their trumbowyg button config to use 'strikethrough' instead of 'del'

### DIFF
--- a/src/trumbowyg.js
+++ b/src/trumbowyg.js
@@ -14,7 +14,7 @@ jQuery.trumbowyg = {
 
             bold: 'Bold',
             italic: 'Italic',
-            strikethrough: 'Stroke',
+            strikethrough: 'Strikethrough',
             underline: 'Underline',
 
             strong: 'Strong',
@@ -89,7 +89,7 @@ Object.defineProperty(jQuery.trumbowyg, 'defaultOptions', {
             ['viewHTML'],
             ['undo', 'redo'], // Only supported in Blink browsers
             ['formatting'],
-            ['strong', 'em', 'del'],
+            ['strong', 'em', 'strikethrough'],
             ['superscript', 'subscript'],
             ['link'],
             ['insertImage'],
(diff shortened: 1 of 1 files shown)


DmitryKmita:
Tested, fix does work!

aleksandrs-ledovskis:
@Alex-D What is your opinion regarding this change? Do you see it eventually getting merged?

src/trumbowyg.js
```diff
@@ -89,7 +89,7 @@ Object.defineProperty(jQuery.trumbowyg, 'defaultOptions', {
             ['viewHTML'],
             ['undo', 'redo'], // Only supported in Blink browsers
             ['formatting'],
-            ['strong', 'em', 'del'],
```
Alex-D: I do not want to change the default semantic tag.

src/trumbowyg.js
```diff
@@ -441,8 +441,7 @@ Object.defineProperty(jQuery.trumbowyg, 'defaultOptions', {
         DEFAULT_SEMANTIC_MAP: {
             'b': 'strong',
             'i': 'em',
-            's': 'del',
-            'strike': 'del',
```
Alex-D: I want to keep `del` if semantic option is set to true.

Alex-D:
I've made some test, and this code do nothing else that we can already do by disabling `semantic`, or customizing the semantic map and use `strikethrough` instead of `del` in `btns` option.

I keep `strike` tag in definition, because it's the tag that browsers use. That's not my choice.

So, sorry but that just broke semantic and solve nothing :/

PR #946, opened 2019-03-30
Missing Slovenian translation

matjaz321:
### Problem
- Missing Slovenian translation
### Solution
- Translated into Slovenian locale
### Additional
- I am a native Slovenian speaker and Microsoft provides translation for Excel/Office/Word etc. I double checked translations so everything should be good to go :wink:


Alex-D:
Thanks! :)

PR #948, opened 2019-04-02
[Plugin] Fix table size selection dropdown from highlighting wrong number of rows

SPLipka:
**Problem:**
When the Trumbowyg container is nested inside a table the table size selection dropdown highlights the wrong number of rows and columns as shown [here](http://jsfiddle.net/splipka/dfgnjayk/7/). When the table variable is defined using .parents() it's selecting the containing table as well as the table created for the dropdown. 

**Solution:**
Using .closest() instead of .parents() narrows the selector to only the table created within the dropdown.


Alex-D:
Nice catch!

Thanks :)

PR #970, opened 2019-05-12
feat: update korean translation for core and plugin

SDuck4:
Changed some words and translate plugins.


Alex-D:
Oh! Thank you from the other side of the Earth! (France)

PR #971, opened 2019-05-14
docs: fix wrong plugin path

SDuck4:
I found some wrong plugin path in docs. And I fixed it.


Alex-D:
Arf, Me? Copy pasting? :-°

Thanks for the fix!

PR #972, opened 2019-05-16
Fixbug: "that is undefined" in upload plugin (Firefox)

pbriet:
Fixing a bug when doing an Ajax Post with trumbowyg upload plugin activated.

`TypeError: that is undefined`  (trumbowyg.upload.js:259:32)

Thanks,

### DIFF
--- a/plugins/upload/trumbowyg.upload.js
+++ b/plugins/upload/trumbowyg.upload.js
@@ -259,8 +259,8 @@
             var originalXhr = $.ajaxSettings.xhr;
             $.ajaxSetup({
                 xhr: function () {
-                    var req = originalXhr(),
-                        that = this;
+                    var req = originalXhr();
+                    var that = this;
                     if (req && typeof req.upload === 'object' && that.progressUpload !== undefined) {
                         req.upload.addEventListener('progress', function (e) {
                             that.progressUpload(e);


Alex-D:
That is really strange since I use this syntax in the core and seems working... I am dubitative about this fix.

pbriet:
Well, I'm no javascript expert. Not sure if there can be some troubles with multiple commands separated by commas, or some weird scope issues.

What I'm sure of however :
- I've had this error on the last version of FF
- With the following files included :   trumbowyg.min.js  and trumbowyg.upload.js
- The error occured when we were doing an ajax call (POST), after retrieving the form data
- This PR solved the issue

Here is what our code looked like : 
````
 $.post(url, new FormData(this),  function (data, status)
        {
             ....
        },
    });
````

We tried to upgrade Trumbowyg to see if it could fix the issue, and it didn't. So it has been tested on 2.17.0

Hope it can help.

Thanks,

Alex-D:
I will investigate. Because, if it's the fix, maybe I need to do this also in the core code.

Alex-D:
I can't reproduce :(

I'm on the latest Firefox (66.0.5) on Windows 10. No error at all on this page: https://alex-d.github.io/Trumbowyg/demos/#plugins-upload

Alex-D:
I close this for now, feel free to give me a repro based on this jsfiddle:
https://jsfiddle.net/AlexandreDemode/p7hb3x6u/

PR #978, opened 2019-05-28
Set ol list-style to decimal

liammccartney:
Hi!

I wanted to piggyback off of #697 and add a `list-style` value for `ol` elements.

I'm currently working in an environment that globally sets `list-style: none` for both `ul` and `ol` elements, so while I'm using trumbowyg's `resetCss` feature the ordered lists still don't have a number before each element.

Thanks!


Alex-D:
Nice catch! Thank you for the fix :)

Issue #982, opened 2019-06-05
Invisible textarea has height of 1px


dontub:
The textarea containing the plain HTML has a height of 1px when it is not visible (so actually it is visible but only 1px of it). This results in a small line between the scroll bar and the bottom border. Is there a reason why height is not set to 0? (I'm using Firefox 67.0.1)

Alex-D:
It's for HTML5 validation purpose.

If you do not need that (no required attribute or so), you can add some CSS like so:

```css
.trumbowyg-box.trumbowyg-editor-visible {
    .trumbowyg-textarea {
		display: none;
    }
}
```

I think it should work ^^

dontub:
Thanks for the quick response. Yes, it works. Just for the record in plain CSS it is:
```css
.trumbowyg-box.trumbowyg-editor-visible .trumbowyg-textarea {
    display: none;
}
```

Actually I can live with the 1px line, though I would suggest to add the following rule to avoid the move of following elements when switching the editor:
```css
.trumbowyg-box.trumbowyg-editor-hidden .trumbowyg-textarea {
    margin-bottom: 1px;
}
```

PR #984, opened 2019-06-06
Avoid move of following elements on editor switch

dontub:
See https://github.com/Alex-D/Trumbowyg/issues/982#issuecomment-499081573.


Alex-D:
Ah OK. I understand :)

Thanks for the fix!

PR #985, opened 2019-06-06
Disable word wrapping in dropdowns

dontub:
This screenshot shows the behavior without this change:
fontfamily-dropdown-screenshot


Alex-D:
We should add this to the core CSS, not the plugin's one.

dontub:
You mean as rule for `.trumbowyg-dropdown`?

Alex-D:
Yep!

Alex-D:
Thanks!

PR #1003, opened 2019-07-22
add some table features

sascha-haenlein:
I added some features to the table plugin and modified two existing functions.
The methods "tableAddRow" and "tableAddColumn" where modified to add there row / column not at the end of the table. Instead the row / column will be added after the selected cell.
Further I implemented two new methods "tableAddColumnLeft" and "tableAddRowBefore" which will add a row / column before the selected cell.

-----------------

add function to add column before selected cell
change function tableAddColumn to add column after selected cell
add function to add row before selected cell
change function tableAddRow to add row after selected cell

### DIFF
--- a/plugins/table/trumbowyg.table.js
+++ b/plugins/table/trumbowyg.table.js
@@ -21,39 +21,49 @@
             en: {
                 table: 'Insert table',
                 tableAddRow: 'Add row',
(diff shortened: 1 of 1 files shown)


Alex-D:
Thanks! :)

PR #1016, opened 2019-08-30
Trigger 'tbwchange' change on viewHTML click

ProbableTrain:
When clicking viewHTML, unclosed tags are automatically closed. This causes the textarea content to change, but currently no event is triggered.

When using Trumbowyg with Angular, this can cause issues as the model is not updated, and consequently the Angular model is not the same as the content shown in viewHTML mode. 

This change triggers the 'tbwchange' event when the viewHTML button is clicked, ensuring that if the browser changes the source, this is reflected.


Alex-D:
Thank you for the fix :)

PR #1025, opened 2019-10-02
Add missing "align" translation

Fyers:


Alex-D:
This translation key is not used anywhere and does not exists in the english translation object here: 

https://github.com/Alex-D/Trumbowyg/blob/8b64a97d7a83cf421d572f6996ec6eda3de33088/src/trumbowyg.js#L1-L57

Issue #1041, opened 2019-10-23
Bug. Deleted text formatting


Cherinho:
On the demo-page (https://alex-d.github.io/Trumbowyg/) and of course if I use it the deleted text formatting doesn't working correclty. I can set text as deleted and button will highlighted. But after I click this button second time(for cleaning formatting) nothing happened and text still have deleted formatting style

Alex-D:
Seems fixed now :)

PR #1043, opened 2019-10-29
fix semantic mapping of del button by tagging it with "s"

ymc-sise:
Fix for https://github.com/Alex-D/Trumbowyg/issues/1041


ymc-sise:
Sorry this is wrong, how can I delete the pull request?

ymc-sise:
Problem is here https://github.com/Alex-D/Trumbowyg/blob/develop/src/trumbowyg.js#L457: Creation of two same newTags "del".

PR #1049, opened 2019-11-08
[Plugin] resizimg updated to use canvas for resize operation

PiemP:
Hi, 

I have updated the resizimg plugin to render it more user friendly and to solve a couple of issue like
* https://github.com/Alex-D/Trumbowyg/issues/1045
* allow to move images inside HTML code when resizimg plugin ins enabled 
....
This is my code that use an HTML canvas to create a zone where resize the image with the jquery-resizable code, and this is a little demo:
 
image

I'm waiting for reviews. Now I'm trying to add some key binding to improve its functionality.

Hope to be useful 

P.S. https://github.com/OrchardCMS/OrchardCore/pull/4490

PR #1062, opened 2019-12-17
[Plugin] resizimg _ precedence fix on first click over image and function wrap

PiemP:
Wrapped ResizeWithCavas function and modified a line of code to avoid strange issue on first click over an image: it fire destroyResizable before canvas is set as focused.

See https://github.com/Alex-D/Trumbowyg/pull/1049

### DIFF
--- a/plugins/resizimg/resize.with.canvas.js
+++ b/plugins/resizimg/resize.with.canvas.js
@@ -4,28 +4,28 @@
     window.ResizeWithCanvas= function() {
 
         //variable to create canvas and save img in resize mode
-        this.resizecanvas = document.createElement('canvas');
+        this.resizeCanvas = document.createElement('canvas');
         //to allow canvas to get focus
-        this.resizecanvas.setAttribute('tabindex','0');
-        this.resizecanvas.id = 'tbwresizeme';
+        this.resizeCanvas.setAttribute('tabindex','0');
+        this.resizeCanvas.id = 'tbwresizeme';
         this.ctx = null;
-        this.resizeimg = null;
+        this.resizeImg = null;
 
         /* jshint unused:vars */
         //function callback to do something when appen something
-        this.beforecanvasreplaced = function (canvas, image){
+        this.beforeCanvasReplaced = function (canvas, image){
 
         };
-        this.presskeyesc = function (obj){
+        this.pressKeyEsc = function (obj){
             obj.reset();
         };
-        this.presskeydelorcanc = function (obj){
-            $(obj.resizecanvas).replaceWith('');
-            obj.resizeimg = null;
+        this.pressKeyDelOrCanc = function (obj){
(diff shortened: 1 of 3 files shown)


plugins/resizimg/resize.with.canvas.js
```diff
@@ -1,212 +1,218 @@
-function ResizeWithCanvas() {
+var ResizeWithCanvas = (function ($) {
```
Alex-D: remove the `var ... =` part, it's not useful here ^^

PiemP: doh! 🙈

Sorry, but I missed a lot of javascript best practice.

Alex-D:
I can see some lint issues. You can use yarn lint to see them and fix them :)

PiemP:
Thank you @Alex-D, 

there are a lot of things new for me. Hope to have fixed all of the lint issues in the right way.

plugins/resizimg/resolveconflict-resizable.js
```diff
@@ -1,4 +1,4 @@
-(function(factory, undefined) {
+(function(factory, undefined, define, require, module) {
```
Alex-D: undefined must be the last argument

plugins/resizimg/resize.with.canvas.js
```diff
@@ -11,10 +11,18 @@
         this.ctx = null;
         this.resizeimg = null;
 
+        /* jshint unused:vars */
         //function callback to do something when appen something
-        this.beforecanvasreplaced = function (canvas, image){};
-        this.presskeyesc = function (obj){};
-        this.presskeydelorcanc = function (obj){};
+        this.beforecanvasreplaced = function (canvas, image){
```
Alex-D: I prefer camelCase names:

`beforeCanvasReplaced`

Also, there are some lint issues. And why this function is empty?

plugins/resizimg/resize.with.canvas.js
```diff
@@ -11,10 +11,18 @@
...
+        this.presskeyesc = function (obj){
+            obj.reset();
+        };
+        this.presskeydelorcanc = function (obj){
+            $(obj.resizecanvas).replaceWith('');
+            obj.resizeimg = null;
+        };
 
         //PRIVATE FUNCTION
         var isfocusednow = false;
```
Alex-D: `isFocusedNow`

plugins/resizimg/resize.with.canvas.js
PiemP: This function is empty because It use to modify the image before it is restored in the HTML of the editor. For example, if you have some filter that allow you to create a version of your image already resized, you could modify the css class or the src of the image to do it. I leave it empty because is the default behavior: restore the image without something of special.

Alex-D:
Thank you :)

PR #1073, opened 2020-01-30
Merge pull request #1073 from gabe565/fontsize-child-tags

### LINKED ISSUE #1072: Font size plugin removes any formatting in range
### Description

If a range contains any sort of text formatting (Nested html tags), then when the font size is changed, all of the formatting is lost. It can be reapplied, but it will be lost if the size is changed again.

### How to reproduce?

1. Type some text with any sort of formatting (Link, bold, italic, strikethrough, changed font, etc)
Screen Shot 2020-01-28 at 5 24 41 PM
```html
<p>Text with a <a href="https://example.com" target="_blank">link</a> <strong>and</strong> <em>some</em> <span style="font-family: &quot;Courier New&quot;, Courier, monospace;">formatting</span> inside the highlight</p>
```

2. Highlight the text and change the font size. All formatting will be lost and the html structure will be replaced with a single `span`.
Screen Shot 2020-01-28 at 5 20 24 PM
```html
<p><span style="font-size: x-large;">Text with a link and some formatting inside the highlight</span></p>
```

gabe565:
This changes how font-size is applied.   
`trumbowyg.execCmd('fontSize', '1')` is used, then any instance of `font[size="1"]` is found and replaced with a span which sets the correct size.

I noticed that spans are left if a user changes the font-size of a selection, then changes the font-size of another selection around that, so any span with an empty style tag is unwrapped to keep the html structure clean while a user is resizing. Thoughts on this part? Do you think that is too dangerous to do?

Edit: This fixes #1072

### DIFF
--- a/plugins/fontsize/trumbowyg.fontsize.js
+++ b/plugins/fontsize/trumbowyg.fontsize.js
@@ -201,16 +201,21 @@
     function setFontSize(trumbowyg, size) {
         trumbowyg.$ed.focus();
         trumbowyg.saveRange();
-        var text = trumbowyg.range.startContainer.parentElement;
-        var selectedText = trumbowyg.getRangeText();
-        if ($(text).html() === selectedText) {
-            $(text).css('font-size', size);
-        } else {
-            trumbowyg.range.deleteContents();
-            var html = '<span style="font-size: ' + size + ';">' + selectedText + '</span>';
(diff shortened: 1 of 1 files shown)


Alex-D:
Oh yeah!

Seems really better than the previous version :)

Thank you!

PR #1076, opened 2020-02-05
Sort plugin translations alphabetically

peteraba:
Because currently it's hard to find translations, especially since they're a different set of languages for each plugins.

Note that I left English first to make the life of translators easier, but I'm happy to change it if needed.


Alex-D:
Thank you for your time, nice cleanup! :)

Issue #1078, opened 2020-02-10
npm WARN trumbowyg@2.19.1 requires a peer of jQuery@>=1.8 but none is installed.


seyfer:
When I install my dependencies or build with webpack I see this message

```
npm WARN trumbowyg@2.19.1 requires a peer of jQuery@>=1.8 but none is installed. You must install peer dependencies yourself.
```

### Description
Provide relevant details according to the context:
  - Linux
  - Latest npm and node

### How to reproduce?
Just run the install of any package with webpack, when trumbowyg@2.19.1 is installed

It is really easy to fix maybe? there is no package named `jQuery`. We have `jquery` installed
`"jquery": "^3.4.1",`
means `package.json` in trumbowyg is wrong. should be
```
  "peerDependencies": {
    "jquery": ">=1.8"
  },
```

Alex-D:
Thank you for the fix :)

andreiradoi:
Could we have this pushed to npm as well? Thanks :)

xdmda:
I have same problem with jQuery@1.8. can anyone help me?

Alex-D:
You just need to install jQuery on the version you want.

```zsh
npm install --save trumbowyg jquery
```

PR #1079, opened 2020-02-10
fix for #1078 - jquery peer warn on packages install

seyfer:
https://github.com/Alex-D/Trumbowyg/issues/1078


Alex-D:
Ahah, nice catch! :)

PR #1082, opened 2020-02-16
feat: add the ability to add classes to tags

holema:
… for h1-h5 an p. The classes which should be added the the paragraphes should be an array in the constructor.

### DIFF
--- a/src/trumbowyg.js
+++ b/src/trumbowyg.js
@@ -1448,15 +1448,9 @@ Object.defineProperty(jQuery.trumbowyg, 'defaultOptions', {
                         }
                         var arr = t.o.tagClasses[param];
                         if (arr) {
-                            $(listId).removeClass();
-                            for (var i = 0; i < arr.length; i++) {
-                                $(listId).removeClass(arr[i]);
-                            }
                             for (var i = 0; i < arr.length; i++) {
                                 $(listId).addClass(arr[i]);
                             }
-
-
                         }
                     } catch (e) {
 
--- a/src/trumbowyg.js
+++ b/src/trumbowyg.js
@@ -77,6 +77,12 @@ Object.defineProperty(jQuery.trumbowyg, 'defaultOptions', {
         imageWidthModalEdit: false,
 
         prefix: 'trumbowyg-',
+// classes for inputs
+        h1Class: null,
+        h2Class: null,
+        h3Class: null,
+        h4Class: null,
+        pClass: null,
 
         semantic: true,
         semanticKeepAttributes: false,
@@ -496,7 +502,7 @@ Object.defineProperty(jQuery.trumbowyg, 'defaultOptions', {
                 return;
             }
 
-            if (typeof(protocol) !== 'string') {
+            if (typeof (protocol) !== 'string') {
                 return 'https://';
             }
             return protocol.replace('://', '') + '://';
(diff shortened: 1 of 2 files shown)


src/trumbowyg.js
```diff
@@ -77,6 +77,12 @@ Object.defineProperty(jQuery.trumbowyg, 'defaultOptions', {
         imageWidthModalEdit: false,
 
         prefix: 'trumbowyg-',
+// classes for inputs
+        h1Class: null,
+        h2Class: null,
+        h3Class: null,
+        h4Class: null,
+        pClass: null,
```
Alex-D: Maybe it will be better to have an object like that:

```
tagClasses: {}
```

Which will allow adding class to more tags :)

src/trumbowyg.js
```diff
@@ -1433,6 +1440,24 @@ Object.defineProperty(jQuery.trumbowyg, 'defaultOptions', {
 
                     t.syncCode();
                     t.semanticCode(false, true);
+                    try {
+                        var listId = window.getSelection().focusNode.parentNode;
+                        var arrayKey = param + 'Class';
```
Alex-D: If we do like I said above, there is no need to build this key.

But, instead, we need to check if the tag has a class in the `tagClasses` object.

src/trumbowyg.js
```diff
@@ -1433,6 +1440,24 @@ Object.defineProperty(jQuery.trumbowyg, 'defaultOptions', {
...
+
+                        var arr = t.o[arrayKey];
+                        if (arr) {
+                            $(listId).removeClass();
```
Alex-D: Prefer just remove the class from the option, not *all* classes, since the user can add a class by other ways

holema: is done ;)

src/trumbowyg.js
holema: is done

src/trumbowyg.js
holema: is done

holema:
hello @Alex-D are you going to merge the request in the next version? if not, I have to use my version in my project, which is not my prefered way ;)

src/trumbowyg.js
Alex-D: I did not understand the purpose of this section.
This line removes all classes, then you remove again each class, and finally re-add them back.
We can give a string with all classes to addClass function, so this can be simplified.

Do not forget that users can add classes using other ways, so we do not want to clean them here. Just add the wanted classes.

holema: hello @Alex-D 
I fixed it by remove the class removing lines.

best regards
Emanuel

holema:
Hello Alex, will you merge this branch in the future?

Alex-D:
I hope to merge it for the 2.22.0 soon, I just need to take some time to test it, and maybe rewrite it a bit :)

Alex-D:
After re-read, I did not understand the value of a systematic addition of a class on an element. I let you add this on your own version, but I will not include it, sorry.

Thank you for your time :)

holema:
The possibility is, that when you use bootstrap for examle, you need to add "h1-responsive" class to the h1 tag, to show it properly on all devices. 
Also it is possible to add some sytling classes to the tags, e.Ex. I use special classes on blockquotes and set a z-depth to show drop shadows.

PR #1097, opened 2020-04-16
Add link popup "URL" to translatable strings

bertyhell:
So people can add a different label if they want to.
I've made it backwards compatible by adding the `|| 'URL'` since no translation file has this string yet.

### DIFF
--- a/src/trumbowyg.js
+++ b/src/trumbowyg.js
@@ -1276,7 +1276,7 @@ Object.defineProperty(jQuery.trumbowyg, 'defaultOptions', {
 
             var options = {
                 url: {
-                    label: 'URL',
+                    label: t.lang.linkUrl || 'URL',
                     required: true,
                     value: url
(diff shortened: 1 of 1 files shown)


src/trumbowyg.js
```diff
@@ -1276,7 +1276,7 @@ Object.defineProperty(jQuery.trumbowyg, 'defaultOptions', {
 
             var options = {
                 url: {
-                    label: 'URL',
+                    label: t.lang.linkUrl || 'URL',
```
Alex-D, line 1279: why not just `t.lang.url`?

bertyhell: because the 'URL' string is used in other locations in your code and if you want to replace those as well, they might collide. eg: image URL

Alex-D: Oh, nice catch :)

Issue #325, opened 2016-07-10
SVG icons not loading


AttitudeMonger:
I downloaded the zipped structure, built it with bower and loaded the `index.html` page in `examples`. The icons are not showing up. 

As per the documentation here: http://alex-d.github.io/Trumbowyg/documentation.html#svg-icons, I added this line of code before `$('#editor').trumbowyg();`:

`$.trumbowyg.svgPath = '/Users/attitude/Desktop/Trumbowyg-develop/dist/ui/icons.svg';`

Same result. How do I get the icons to load correctly? Or is it that because they will be loaded by XHR, they will show up only in a server-client environment, not if the page is loaded in the browser locally?

Alex-D:
You need to be in HTTP(S)

AttitudeMonger:
Okay, I will try it. Can you give me an idea about the directory structure? Which files are important? I can see there are three - the `.svg` file and `.css` file in `dist/ui`, and `.js` file in `dist`. Is that all? Do I need to include the directory which contains lots of `svg` image files as well? If so, what should be its directory level with respect to these three files?

Alex-D:
You should only use files from [`dist`](https://github.com/Alex-D/Trumbowyg/tree/master/dist) folder.
The only constraint by default is to [have SVG next to CSS](https://github.com/Alex-D/Trumbowyg/tree/master/dist/ui).

SuN-80:
Same thing here. I added the path to svg sprite (I used dist/ui/icons.svg), but it remains undefined int he DOM. What's the problem please?

AttitudeMonger:
Thanks. I got it working. However, the color buttons are not coming, even though I have included both the `js` and `css` files for the colors. This is the sample code I am trying

```
$('#editor')
            .trumbowyg({
                btnsAdd: ['foreColor', 'backColor']
            });
```

Alex-D:
I can't help you without context.

- The option is here: http://alex-d.github.io/Trumbowyg/documentation.html#svg-icons
- You can add SVG directly in your page without AJAX like this: https://github.com/Alex-D/Trumbowyg/blob/gh-pages/documentation.html#L16 with all [icon.svg](https://github.com/Alex-D/Trumbowyg/blob/master/dist/ui/icons.svg) content.

Issue #253, opened 2016-04-11
SCV icons path problem


akupiec:
### Informations
It's quite problematic to load icons through javaScript.
The only reason i decided to user trumbowy was clean way to replace all css and icons.

So pleas add easy way to give relative path for icons inside *.scss. (for example variable with path like it was till 2.0.0-beta.8)
(it it's already possible please mention this in documentation)

Moreover not setting `svnPath `is throwing error in console.

**Browser + version**: chrome/ie10
**OS**: win
**Resolution**: 2.0.0+

### How to reproduce the bug?
try to use Trumbowyg without setting path to icons through js.

RespectTheCoder:
Use `svgPath`, also see #256.

Alex-D:
I've just added a global `$.trumbowyg.svgPath` option which can be set to false to disable SVG js loading.

I can't just hide network error and I can't put SVG management in CSS due to actual support of SVG features sucks, e.g. I

briansanchez:
this works, thank you Alex.
//2019-08-16 angularjs fix, when minimizing will help to avoid >> You must define svgPath: https://goo.gl/CfTY9U
$.trumbowyg.svgPath = '/static/bower_components/trumbowyg/dist/ui/icons.svg';

PR #1102, opened 2020-04-25
feat: new option svgSideLoad = false

keithharvey:
Allows the svgSideLoad option (default true) to force use of the svgPath in the href of the svg. The svgSideLoad option allows skipping dom insertion and makes it explicit how the svg is being handled since the default behavior is surprising when passing in an svgPath (maybe I was not the only one confused by this? https://github.com/Alex-D/Trumbowyg/issues/325).

## Context ##
The angularjs application I was working on was rendering trumbowyg inside a dynamic tab. Switching tabs would cause the dom reference technique to fail whenever it re-rendered via client side rendering, tab switches. I saw the option for svgPath and immediately assumed I could just pass in an explicit url. Instead I found the current page URL always and the current implementation, which exists to allow styling of the svg (https://github.com/Alex-D/Trumbowyg/issues/253#issuecomment-217642450). Editing the svg html and setting the href to the svgPath I passed into to trumbowyg revealed that just letting the browser do it solved my re-rendering problem.

In my case I'm stuck with a legacy app and just want them to work so the browser caching them is fine.

### DIFF
--- a/src/trumbowyg.js
+++ b/src/trumbowyg.js
@@ -112,7 +112,8 @@ Object.defineProperty(jQuery.trumbowyg, 'defaultOptions', {
         plugins: {},
         urlProtocol: false,
         minimalLinks: false,
-        defaultLinkTarget: undefined
+        defaultLinkTarget: undefined,
+        svgSideLoad: true
     },
     writable: false,
     enumerable: true,
@@ -214,11 +215,19 @@ Object.defineProperty(jQuery.trumbowyg, 'defaultOptions', {
 
         t.hideButtonTexts = $trumbowyg.hideButtonTexts != null ? $trumbowyg.hideButtonTexts : options.hideButtonTexts;
 
+        // Defaults Options
+        t.o = $.extend(true, {}, $trumbowyg.defaultOptions, options);
+
         // SVG path
         var svgPathOption = $trumbowyg.svgPath != null ? $trumbowyg.svgPath : options.svgPath;
         t.hasSvg = svgPathOption !== false;
-        t.svgPath = !!t.doc.querySelector('base') ? window.location.href.split('#')[0] : '';
-        if ($('#' + trumbowygIconsId, t.doc).length === 0 && svgPathOption !== false) {
+        if (!options.svgSideLoad && svgPathOption == null) {
+          console.warn('You must define svgPath: https://goo.gl/CfTY9U'); // jshint ignore:line
+        }
+        var baseHref = !!t.doc.querySelector('base') ? window.location.href.split('#')[0] : ''
+        t.svgPath = options.svgSideLoad ? baseHref : svgPathOption;
+        
+        if (options.svgSideLoad && $('#' + trumbowygIconsId, t.doc).length === 0 && svgPathOption !== false) {
             if (svgPathOption == null) {
                 // Hack to get svgPathOption based on trumbowyg.js path
                 var scriptElements = document.getElementsByTagName('script');
@@ -407,8 +416,6 @@ Object.defineProperty(jQuery.trumbowyg, 'defaultOptions', {
             }
         };
 
-        // Defaults Options
-        t.o = $.extend(true, {}, $trumbowyg.defaultOptions, options);
         if (!t.o.hasOwnProperty('imgDblClickHandler')) {
             t.o.imgDblClickHandler = t.getDefaultImgDblClickHandler();
         }


Alex-D:
Hey!

I have read, and re-read your PR, and I'm sorry but I did not understand the need.

Why can you just not import SVG in HTML of your tab? Maybe before Trumbowyg init. Since I did not understand the point of your code, I cannot merge it. Also I cannot write documentation for it.

I have already an idea to avoid all those issues in v3, because SVG loading issues are annoying.

keithharvey:
> Why can you just not import SVG in HTML of your tab?

You totally can. If you inject it into your layout it does work. That is an option. The point of this was that `svgPath: '/icons.svg'`, by default, I would expect to load that from the server and put it in the href for the svg. That just works for me, so adding an option to both make explicit the side loading and to allow you to just skip it is an improvement on the API. Not strictly necessary, just nice. And I could update the docs, just wanted to run it by you before going through the trouble.

src/trumbowyg.js
```diff
@@ -153,10 +153,16 @@ jQuery.trumbowyg = {
         t.hideButtonTexts = $.trumbowyg.hideButtonTexts != null ? $.trumbowyg.hideButtonTexts : options.hideButtonTexts;
 
         // SVG path
-        var svgPathOption = $.trumbowyg.svgPath != null ? $.trumbowyg.svgPath : options.svgPath;
+        var svgPathOption = $trumbowyg.svgPath != null ? $trumbowyg.svgPath : options.svgPath;
```
Alex-D, line 156: I think this line should remain as is.

src/trumbowyg.js
```diff
@@ -153,10 +153,16 @@ jQuery.trumbowyg = {
...
         t.hasSvg = svgPathOption !== false;
-        t.svgPath = !!t.doc.querySelector('base') ? window.location.href.split('#')[0] : '';
-        if ($('#' + trumbowygIconsId, t.doc).length === 0 && svgPathOption !== false) {
+        var sideLoad = options.svgSideLoad === false ? false : true;
```
Alex-D, line 158: A Cleaner way is: 

    var sideLoad = options.svgSideLoad !== false;

keithharvey: I was just trying to assure that it was in fact a boolean and not relying on the truthiness downstream. You're right about the simplification, thanks.

keithharvey:
Had a few force pushes messed up there but updated with your feedback and rolled it into one commit. If you would you like me to just use `options.svgSideLoad` instead of the variable just let me know.

Alex-D:
I'm thinking that it should be better to init the option in the default options object, to true, allowing us to remove that check ^^

keithharvey:
Went ahead and did that. I did have to move t.o initialization above the svg initialization because the defaults weren't quite pulled in yet, but that all seems to work.

Alex-D:
Looks good
I will take some time to test it in real context + write the doc ASAP :)

Thank you!

keithharvey:
Thank you! I really appreciate your thoughtfulness.
---
So, last night I finished up the changes you asked for and then went about looking for another way to wrap angularjs's rendering lifecycle so that maybe my changes weren't necessary and I found one. Turns out switching to the techniques used [here](https://github.com/jutaz/angular-trumbowyg) instead of the ones used [here](https://github.com/lizardK/trumbowyg-ng) (linked in your docs) eliminated the svg rendering issues. With that in mind, I'm not sure I would merge this pull request until someone else +1'd this for a real-world use case. The docs and API transparency points aren't nothing, but it is a tad more complexity for 0 functional gain.
----
A separate issue with both wrappers involves the html null check found in trumbowyg.

[Copied here for clarity](https://github.com/Alex-D/Trumbowyg/blob/v2.21.0/src/trumbowyg.js#L1102)
```
            if (html != null) {
                t.$ta.val(html);
                t.syncCode(true);
                t.$c.trigger('tbwchange');
                return t;
            }
```
In angular-trumbowyg, I simply had to change $render to include this `if` to allow my application to push null state into trumbowyg.
```
          if (this.ngModel.$modelValue !== undefined && this.ngModel.$modelValue == null) {
            value = '';
          } else {
            value = this.ngModel.$modelValue;
          }
          element.trumbowyg('html', value);
```
A dropdown changed the locale key used for data binding. The new key is initialized, but null, which I would expect calling trumbowyg('html', null) to result in an empty plugin..

keithharvey:
@Alex-D My colleague added a fix we found. You can reproduce it by adding a new 3x3 table using the table plugin, then clicking off to the side of the top left cell. The wrapper fires but doesn't know about the table yet, so ends up wrapping each td in a p tag, which chrome redraws for you when it discovers how invalid it is.

Alex-D:
1/ I think it should be in another PR
2/ I think it's due to missing `syncCode` in table plugin, so this is not a fix but a workaround ^^'

I need to take the time to look at the original option, I was working on something else + computer issues :/

keithharvey:
2/ Ok, I force pushed it from this repo. Yeah I couldn't find the execution path that SHOULD call `syncCode`

ydubey:
Hi,

I am struggling with what I think is because of the URL in svg > path element. I am using angular with core-ui. The icons do not show when editor component is inside a tab. Looking at the network traffil, there is a http get request to the current pages url which is returned with 404. I suspect, this is when the svg icons are being rendered.

I dont have much context of how svg is being loaded, but reading the description I think svgSideLoad is the solution.

I tried using latest 2.22.0, with svgSideLoad: false, but that doesnt seem to make any difference to the url on svg > path element.

`
    $("#templateEditor").trumbowyg({
      lang: 'en',
      svgSideLoad: false,
      svgPath: '/assets/icon/icons.svg',
      removeformatPasted: true
    });
`
Am I missing something?

Alex-D:
Yes, sorry, I've renamed the option ^^
It's `svgAbsoluteUseHref` now :)

ydubey:
Hi,

Thanks for the reply. I tried both
`
svgAbsoluteUseHref: false
`
and

`
svgAbsoluteUseHref: '/assets/icon/icons.svg'
`
none seemed to make any difference to the href in svg > use. Is this change part of 2.22.0 version?

Thanks

Alex-D:
Use `svgAbsoluteUseHref: true`

ydubey:
Hi sorry so say, even that doesn't work. Let me try to create a sample angular project for you to debug and see whats going on.

Thanks for your help.

Alex-D:
That's not an instance option. It's a global option, like `svgPath`.

```javascript
$.trumbowyg.svgPath = '/assets/icon/icons.svg`
$.trumbowyg.svgAbsoluteUseHref = true
```

This will apply to all Trumbowyg instances.

ydubey:
Thanks, that worked.

PR #1112, opened 2020-05-03
Update ko.js

opdev1004:
For insertImage, changing 그림 to 이미지.

Because 그림 has more meaning of drawn piece of art.

Just making sure the word to be more familiar to majority of people as 이미지 is more being used for web editors. 

Added translation for insertVideo.


Alex-D:
Thank you :)

PR #1113, opened 2020-05-04
resizimg IE 11 fix

alex-dolzhok:
An amendment for Unspecified error in ie11 console. For resizeimg plugin.

### DIFF
--- a/plugins/resizimg/trumbowyg.resizimg.js
+++ b/plugins/resizimg/trumbowyg.resizimg.js
@@ -32,6 +32,7 @@
         // PRIVATE FUNCTION
         var focusedNow = false;
         var isCursorSeResize = false;
+        var attachedToDom = false;
 
         // calculate offset to change mouse over square in the canvas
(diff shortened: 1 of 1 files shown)


plugins/resizimg/trumbowyg.resizimg.js
```diff
@@ -14,6 +14,7 @@
     var ResizeWithCanvas = function () {
         // variable to create canvas and save img in resize mode
         this.resizeCanvas = document.createElement('canvas');
+        document.head.appendChild(this.resizeCanvas);
```
Alex-D, line 17: Why in head?
This should be removed from DOM at some point I think

alex-dolzhok: add / remove canvas on setup / reset functions call

Alex-D:
As I cannot reproduce the bug, I do not understand the need behind this fix, so I close this.

PR #1114, opened 2020-05-05
Keep some styling.

alex-dolzhok:
Keep background-color, color, font-size while pasting content.
Undo changes from previous commit.


Alex-D:
The goal of cleanpaste is to remove all of that. So I will not merge since it can be considered as a bug for other users.

alex-dolzhok:
Thanks for review anyway!

PR #1116, opened 2020-05-08
fix(resizimg): clean code and add missing syncCode calls

PiemP:
changed code to avoid error after last jquery updates because replaceWith function now doesn't work with empty string but work with DOM elements or strings that represents DOM elements.

### DIFF
--- a/plugins/resizimg/trumbowyg.resizimg.js
+++ b/plugins/resizimg/trumbowyg.resizimg.js
@@ -1,4 +1,4 @@
-;(function ($) {
+; (function ($) {
     'use strict';
 
     var defaultOptions = {
@@ -26,7 +26,7 @@
         this.pressBackspaceOrDelete = function (obj) {
             $(obj.resizeCanvas).remove();
             obj.resizeImg = null;
-			if(trumbowyg !== null)
+            if (trumbowyg !== null)
                 trumbowyg.syncCode();
         };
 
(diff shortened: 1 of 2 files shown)


plugins/resizimg/trumbowyg.resizimg.js
```diff
@@ -191,25 +193,26 @@
...
-        // clean html code
-        trumbowyg.$ed.find('canvas.resizable')
-            .resizable('destroy')
-            .off('mousedown', preventDefault)
-            .removeClass('resizable');
-
-        resizeWithCanvas.reset();
-
-        trumbowyg.syncCode();
-    }
```
Alex-D, line 207: What's the point of moving that function into the resizimg init function?

PiemP: Hi @Alex-D , I have modified the constructor to pass an instance of trumbowyg to fix an issue that occour when you start resize an image, press canc or del to delete the image and save your text. In the previous code, if you do this actions, the image is not removed from the html code in trumbowyg and you can see it again on the next edit.

I have moved the destroyResizable function too because it uses the instance of ResizeWithCanvas inside it.

PiemP:
I'm sorry, about my need to add more bugfixing or functionality to this PR but some functionality depends on the changes made in this PR: in specific the changes `ResizeWithCanvas` constructor that pass an instance of trumbowyg to this object.

If you believe that I have to split this PR, I trying to do my best to follow your requirement.

lazybird:
Hi @Alex-D and @PiemP ,
I'm using Trumbowyg on my project - thanks for you great work !

Do you guys think we can move forward with this PR ? I'm confronted with the issue described #1158 and I think this PR solved that issue.

If this is merged, do you think it worth building and tagging a new release ? I would love so.

PiemP:
> Hi @Alex-D and @PiemP ,
> I'm using Trumbowyg on my project - thanks for you great work !
> 
> Do you guys think we can move forward with this PR ? I'm confronted with the issue described #1158 and I think this PR solved that issue.
> 
> If this is merged, do you think it worth building and tagging a new release ? I would love so.

Sorry @lazybird, 

but this things doesn't depend by me...for my side the PR is ready, I use it in my application and, for now, I don't have problems. Yes, I have made a lot of confusion in this PR because I solve a lot of issue, but merge this PR in the main code doesn't depend by me.

About the code review open I don't understand if exist something not correct in it. If I can do something to push this PR, I'm here.

Alex-D:
Thank you for your fix/refactor @PiemP :)

PR #1169, opened 2020-10-22
imageFloatModalEdit feature

kckrinke:
Following the same pattern as `imageWidthModalEdit`, users can now enable `imageFloatModalEdit` which will allow them to select from: "none", "left" and "right" options when inserting / editing images inline.


Alex-D:
Hi, thank you for your work!
I'm sorry but I do not want to a add any `select` in v2 since it adds a lot of code to the core.

Your code is really clean by the way, I just want to keep it light :)

PR #1170, opened 2020-10-22
Select wp media library

kckrinke:
Implements rudimentary support for WordPress Media Library if found to be present (`window.wp.media` exists).

Simply adds a button to the image modal dialog which opens the WP Media Library selector and then pre-populates the `url` and `description` fields of the image modal dialog.

When implementing Trumbowyg in a WordPress plugin or whatnot, be sure to `wp_enqueue_media();` somewhere sane before loading the Trumbowyg dist files.

There is no actual WordPress plugin made for Trumbowyg to go along with this feature. Implementation within WordPress is left to the developer interested in this feature.

If `window.wp.media` does not exist, this feature does nothing and everything works as-is and without the `Select WP Media Library` button in the image modal dialog.

### DIFF
--- a/src/trumbowyg.js
+++ b/src/trumbowyg.js
@@ -120,6 +120,37 @@ Object.defineProperty(jQuery.trumbowyg, 'defaultOptions', {
     configurable: false
 });
 
+var TrumbowygMediaLibrary = null;
+function TrumbowygSelectMedia() {
+  'use strict';
+  if (TrumbowygMediaLibrary) {
+    TrumbowygMediaLibrary.open();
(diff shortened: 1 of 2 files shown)


kckrinke:
I've implemented a rudimentary WP plugin which demonstrates Trumbowyg in WordPress.

https://github.com/kckrinke/trumbowyg-editor

Alex-D:
I keep that in mind for v3. But for now, it's really a heavy addition in v2, so I will not merge, sorry.

By the way, thank you for your PR :)

PR #1208, opened 2021-03-31
Don't turn off removeformatPasted if not needed

### LINKED ISSUE #1207: AllowTagsFromPaste disables removeformatPasted even if disabled
AllowTagsFromPaste plugin forces `removeformatPasted` to `false` before checking if itself is enabled.

https://github.com/Alex-D/Trumbowyg/blob/1d4fbb9ad4d05219d61a032d13ed8baf3e2c0a6e/plugins/allowtagsfrompaste/trumbowyg.allowtagsfrompaste.js#L133-L137

`removeformatPasted` should not be forced to `false` if AllowTagsFromPaste plugin is disabled in config.

TFleury:
Fixes #1207


Alex-D:
Oh, nice catch! :)

PR #1228, opened 2021-06-29
feat: add custom field types to open modal insert

nzmattman:
I have added an option for creating custom HTML form types by using a call back function as the type.
- Created the callback function
- Adjusted the HTML build process to check for the function type and inject that HTML, otherwise use the standard HTML markup
- adjusted the validation rules to target new markup

I have also adjusted the modal css to allow for this integration to work better;
- changed to use flexbox over absolute positioning
- added styling support for select lists and textareas (rather than just input fields)


Alex-D:
This idea is brilliant! :)

But, I can't merge this since it's full of breaking changes :(

Trumbowyg 2 partially supports IE11, with this, it will completely break the support of all old browsers, which means that I need to push that in a new major version...

Can we make that work without flexbox, and without ES2015+ things?

src/trumbowyg.js
```diff
@@ -1668,11 +1668,30 @@ Object.defineProperty(jQuery.trumbowyg, 'defaultOptions', {
...
+                          type="${field.type || 'text'}"
+                          name="${n}"
+                          id="form-${fieldName}"
+                          ${attr ? ' '+attr : ''}
+                          ${field.type === 'checkbox' && field.value ? ' checked="checked"' : ''}
+                          value="${field.value || ''}"
+                        />
+                      </div>
+                    </div>
+                  `;
```
Alex-D, line 1692: Not supported in older browsers :/
Also, add some bytes (I did not check how many)

src/trumbowyg.js
```diff
@@ -1684,8 +1703,8 @@ Object.defineProperty(jQuery.trumbowyg, 'defaultOptions', {
                     $.each(fields, function (fieldName, field) {
                         var n = field.name || fieldName;
 
-                        var $field = $('input[name="' + n + '"]', $form),
-                            inputType = $field.attr('type');
+                        var $field = $(':input[name="' + n + '"]', $form),
+                            inputType = $field[0].nodeName;
```
Alex-D, line 1707: This will return input, textarea, select, but will not return checkbox nor radio, am I right?

src/ui/sass/trumbowyg.scss
```diff
@@ -347,80 +347,88 @@ $slow-transition-duration: 300ms !default;
...
             transition: all 150ms;
+            flex: 0 0 150px;
+            border-right: 1px solid #DEDEDE;
+            padding: 0 7px;
+            background-color: lighten($light-color, 5%);
 
-            span {
-                display: block;
+            label {
+                cursor: pointer;
```
Alex-D, line 368: Why putting a cursor pointer on a label?

Alex-D (changes requested):

nzmattman:
Ahh right, bloody IE!

Yip, I will change it to not use flexbox.  I haven't had to code anything for IE11 for a long time, but will do

src/trumbowyg.js
nzmattman: No problems, will update.
I have found this a lot cleaner to read than the old way, but if you still support it, easy as

src/ui/sass/trumbowyg.scss
nzmattman: this is a force of habit, it shows that the label is clickable, which really it is.
happy to remove it if not necessary

src/trumbowyg.js
nzmattman: i just did up a quick jsfiddle
https://jsfiddle.net/46Le9quy/
it looks like it does support most fields

nzmattman:
Ok, I have done a few updates for this.

I am not really happy with how the example in the docs looks, but the general code is there

src/trumbowyg.js
Alex-D: OK... just select is `select-one` or `select-multiple` when adding attribute `multiple`, idk if its intuitive

src/ui/sass/trumbowyg.scss
Alex-D: I prefer to keep things as default :)

Alex-D:
I will check that on old browsers in the next few days, thank you for your reactivity! :)

src/trumbowyg.js
Alex-D: The difference between the code here and your demo is `.type` vs `.nodeName`, I will fix that :)

Alex-D (approved):

Alex-D:
Thank you for this PR :)

Alex-D:
I've fixed some bugs on the develop branch directly :)

nzmattman:
You are a champion, thanks Alex.
I didn't have time on Friday to look at this and was going to fix up the small things today.

PR #1238, opened 2021-07-27
Fixes "Cannot read property '1' of null"

alegag:
Our Sentry often report error on color plugin "Cannot read property '1' of null".

Seems to occure when invalid `rgb` is passed as color to `colorToHex` function.

To avoid the issue, for the result of `rgb.match()`, if `null`, color was not decoded correctly, so returns `transparent` instead of using `null` as an array.


Alex-D:
Thank you for the fix :)

Issue #1024, opened 2019-09-25
Formulas not editable after ViewHTML


step4:
I have created a Stackblitz with the following problem:

After adding all necessary files for trumbowyg and the MathML plugin everything seems to work fine.
But if you create a formula like `\sum` and switch to ShowHTML mode and then switch back, you cannot edit the formula anymore.

Stackblitz Link: https://stackblitz.com/edit/js-gxp17p

Alex-D:
Seems that event listeners are lost while we switch between views :/

atiabjobayer:
Anyone found a solve yet?

Alex-D:
Fixed in develop branch :)

PR #1244, opened 2021-08-15
Fix: MathML containers editable after ViewHTML

Alolu:
MathML containers cannot be edited through the modal when `ViewHTML` is switched off and on as explained in issue #1024

Now, the plugin will search for MathML containers with the class `mathMlContainer` on `init` and add `onclick` listeners to them.


Alex-D:
Thank you! :)

PR #1268, opened 2021-11-02
+ Add sourcemaps

timint:


gulpfile.js
```diff
@@ -4,7 +4,8 @@
 const gulp = require('gulp'),
     del = require('del'),
     vinylPaths = require('vinyl-paths'),
-    $ = require('gulp-load-plugins')();
+    $ = require('gulp-load-plugins')(),
+    sourcemaps = require('gulp-sourcemaps'),
```
Alex-D, line 8: Why not `$.sourcemaps`?

package.json
```diff
@@ -25,6 +25,7 @@
     "gulp-rename": "2.0.0",
     "gulp-sass": "4.1.0",
     "gulp-size": "3.0.0",
+    "gulp-sourcemaps": "^3.0.0",
```
Alex-D, line 28: I prefer fixed versions for dev/build things

Note: I know that this is the job of the lockfile and everything, but since some dependencies are old or unmaintained, and some are moving, I prefer set the version, and avoid issues with npm install / update.

Alex-D (changes requested):

Alex-D:
Also, can you rework commits to match something more like Karma commit convention?
http://karma-runner.github.io/6.3/dev/git-commit-msg.html

Thank you :)

gulpfile.js
timint: I am totally new to Gulp although I like it very much. I'm a PHP developer so bare with me :)

timint:
> Also, can you rework commits to match something more like Karma commit convention? http://karma-runner.github.io/6.3/dev/git-commit-msg.html
> 
> Thank you :)

I am unfamiliar with Karma and brain muscle is used to these changelog list notations:
+ added
- removed
* changed
! critical

But if you are asking for `fix: Fixed this that` rather than `* Fixed this that` I can do that.

timint:
You know you can cherry pick my commits and adjust them any way you want them.

package.json
timint: Ok, this one was added automatically by npm. I will set it to "3.0.0"

Alex-D:
Thank you :)

PR #1269, opened 2021-11-02
CSS renovation

timint:
A bunch of fixes, adjustments, and improvements to CSS. See list of commits

### DIFF
--- a/src/ui/sass/trumbowyg.scss
+++ b/src/ui/sass/trumbowyg.scss
@@ -1,3 +1,7 @@
+$button-size: 35px;
+$icon-size: 17px;
+$pane-spacing: 5px;
+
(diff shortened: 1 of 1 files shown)


Alex-D:
I see a lot of changes here in the final style, and it is not just a refactor, so I need to test that deeper.

timint:
Sure, branch it, test it. See how you like it. 👍
I am better at CSS than i am at gulp. So I'm sure you will find some nice upgrades. 🙂

There is also a possibility to use CSS variables today `var(--trumbowyg-this-that)` if you would like them defined on the client side.
https://caniuse.com/?search=css%20variables

Alex-D:
Partially merged, thank you :)

PR #1270, opened 2021-11-02
* Fix duplicate banner comments in minified files

timint:
(cherry picked from commit 508ec4493df3dd2ba630b2b2c39da45b4a0aec1a)

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -70,7 +70,11 @@ const scripts = gulp.series(testScripts, function scripts() {
         .pipe(gulp.dest('dist/'))
         .pipe($.size({title: 'trumbowyg.js'}))
         .pipe($.rename({suffix: '.min'}))
-        .pipe($.terser())
+        .pipe($.terser({
+            format: {
+                comments: false
+            }
+        }))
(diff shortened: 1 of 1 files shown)


Alex-D:
What's the issue fixed? I did not get why this is the good way to fix it, instead of removing comments from source files

timint:
If I remember right from back in august the minified files had duplicate txt banners. I can't confirm that the release is that way today. But I had those when running `npm run build`. I think the problem was that you applied the banner for the distributed versions. Then applied it again for the minified versions.

```
/** Trumbowyg v2.25.1 - A lightweight WYSIWYG editor - alex-d.github.io/Trumbowyg - License MIT - Author : Alexandre Demode (Alex-D) / alex-d.fr */
/** Trumbowyg v2.25.1 - A lightweight WYSIWYG editor - alex-d.github.io/Trumbowyg - License MIT - Author : Alexandre Demode (Alex-D) / alex-d.fr */
...
```

Alex-D:
I've done that instead:
https://github.com/Alex-D/Trumbowyg/commit/055a4171e17ca6b6d60751d4558fd5ebd8ccd270

Seems more clean to me :)

PR #1274, opened 2021-11-21
fix: Prevent Firefox from nesting multiple spans on font-size changes

### LINKED ISSUE #1252: Font-size is only updated once, then is no more updated
It seemed that the span tag is not removed when the font size of an element is updated more than one time

font-size issue

aeyoll:
Fix #1252


Alex-D:
Thank you :)

PR #1278, opened 2021-12-14
Added Russian and Belarusian translations for plugins

inavitski-sc:
Hi! As a native speaker of Belarusian and Russian, I would like to contribute and add missing translations for these languages.


Alex-D (approved):
Thank you :)

PR #1279, opened 2021-12-17
Fixed display of svg icons when using the base tag

inavitski-sc:
The URL query string must be included in `xlink:href` references to inline SVG elements when using `<base>` tag. Otherwise, the SVG icons are not displayed.


Alex-D:
Thank you :)

PR #1314, opened 2022-08-17
Small toolbar fixes

urosg80:
Fix dropdown menu position in toolbar when toolbar is set to fixed position and prevent an error when updating toolbar buttons and a tag is not correctly defined.

### DIFF
--- a/src/trumbowyg.js
+++ b/src/trumbowyg.js
@@ -1115,7 +1115,7 @@ Object.defineProperty(jQuery.trumbowyg, 'defaultOptions', {
                 $btn.addClass(prefix + 'active');
 
                 $dropdown.css({
-                    position: 'absolute',
+                    position: t.$btnPane.css('position') === "fixed" ? 'fixed' : 'absolute', // change dropdown menu position to fixed if the toolbar is also fixed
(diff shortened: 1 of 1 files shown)


src/trumbowyg.js
```diff
@@ -1914,6 +1914,7 @@ Object.defineProperty(jQuery.trumbowyg, 'defaultOptions', {
             t.clearButtonPaneStatus();
 
             $.each(tags, function (i, tag) {
+                if (typeof tag === "undefined") return; // in the unlikely event that the tag is not defined skip it as to not raise an error
```
Alex-D, line 1917: I think this should be fixed here instead: 
https://github.com/urosg80/Trumbowyg/blob/d2a066b9eac68f8a34513e620373a28696ae14c2/src/trumbowyg.js#L1975

Alex-D (changes requested):

src/trumbowyg.js
urosg80: Seems so yes, but since I had this fixed present in version 2.21.0 which I used before maybe it is not pertinent anymore so I will first test and review without fix in the latest version and report back.

Alex-D:
Idk if it is steel relevant :/

PR #1318, opened 2022-08-18
Slovenian localization across plugins

urosg80:
Added / modified Slovenian localization across all plugins - should this not be OK (all plugins in one PR) let me know and I will break it apart per plugin.


Alex-D:
Thank you :)

PR #1319, opened 2022-08-18
History plugin events - clean on destroy

urosg80:
Updated assignment / designment of tbwinit and tbwchange events when editor loaded / unloaded dynamically more than once on a page.

No adverse functionality was observed with original code but to be more memory friendly ... ;)


Alex-D (approved):

Alex-D:
Thank you for the fix :)

PR #1321, opened 2022-08-19
Fixed Dutch translation of fontsize plugin

tschoonen:
We recently started using Trumbowyg as an HTML editor and noticed some mistakes in the Dutch translation of the fontsize plugin.


Alex-D:
Thank you )

PR #1326, opened 2022-08-30
Update trumbowyg.js - add predefined targets for anchor elements

coelhudo:
In the current version, when a link is inserted, that's the modal dialog that shows up:

image

The `target` is a text field. However, the `target` attribute [only accepts four keywords](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#attr-target): `_self`, `_blank`, `_parent`, and `_top`. 

I changed the code to use a drop-down menu instead of a text field, so the users don't need to type the value.

image

image

### DIFF
--- a/src/trumbowyg.js
+++ b/src/trumbowyg.js
@@ -115,7 +115,7 @@ Object.defineProperty(jQuery.trumbowyg, 'defaultOptions', {
 
         urlProtocol: false,
         minimalLinks: false,
-        defaultLinkTarget: undefined,
+        defaultLinkTarget: '_self',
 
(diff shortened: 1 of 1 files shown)


Alex-D:
Thank you! :)
I will review that soon, seems good, but I need to try it for real, and maybe make option labels translated

Alex-D:
Merged:
- 2cf579774ced60bc0abc00ef16b484958c037902
- 63b64fae8f838709dfba4ea4e063beb5659fe081 (some additions/fix)
- 
Thank you :)

Issue #1243, opened 2021-08-13
Trumbowyg Giphy Plugin is broken due to mixed result set from Giphy API


synth:
We have noticed the Giphy plugin is broken with some Giphy API responses. By broken, I mean that no images load and I see in the js console below. In the case below, you can see it immediately without doing a search, but it also repeats for some searches, but not all. It seems the Giphy API response does not always include a `downsized` entry in the json payload. 

<img width="1653" alt="Screen Shot 2021-08-13 at 3 38 52 PM" src="https://user-images.githubusercontent.com/10546/129424548-6a99255d-0414-449d-9052-c4f8f5a64ed9.png">

When testing locally, I can see the api response: 

```json
{
   "original":{
      "height":"640",
      "width":"640",
      "size":"15683861",
      "url":"https://media4.giphy.com/media/YRtLgsajXrz1FNJ6oy/giphy.gif?cid=4a2a0f59efjstu3w6te4wcmmmcqqojmg43s28v7xz14mzin9&rid=giphy.gif&ct=g",
      "mp4_size":"2713472",
      "mp4":"https://media4.giphy.com/media/YRtLgsajXrz1FNJ6oy/giphy.mp4?cid=4a2a0f59efjstu3w6te4wcmmmcqqojmg43s28v7xz14mzin9&rid=giphy.mp4&ct=g",
      "frames":"277",
      "hash":"86271fd2c792723e7d72b0f092c9766e"
   },
   "downsized_large":{
      "height":"318",
      "width":"318",
      "size":"5304292",
      "url":"https://media4.giphy.com/media/YRtLgsajXrz1FNJ6oy/giphy-downsized-large.gif?cid=4a2a0f59efjstu3w6te4wcmmmcqqojmg43s28v7xz14mzin9&rid=giphy-downsized-large.gif&ct=g"
   },
   "downsized_medium":{
      "height":"279",
      "width":"279",
      "size":"3495979",
      "url":"https://media4.giphy.com/media/YRtLgsajXrz1FNJ6oy/giphy-downsized-medium.gif?cid=4a2a0f59efjstu3w6te4wcmmmcqqojmg43s28v7xz14mzin9&rid=giphy-downsized-medium.gif&ct=g"
   },
   "fixed_height":{
      "height":"200",
      "width":"200",
      "size":"4725830",
[...]

PR #1339, opened 2022-10-28
Giphy plugin - fixes & enhancements

TwilightCoder:
@Alex-D Thank you for the giphy plugin. Here are some quick fixes and enhancements we've performed that seem worth merging upstream. Please let us know what you think.

Fixes:
- Image fallback: Uses fallback image variant when fetching from API search results - `images.downsized_medium` along with the current `images.downsized`. `images.downsized` attribute often seems to be empty, where `images.downsized_medium` is present [bda5058cfb881ae497bda90c30ca0b419a504c14].
- CSP issue: Moves image `onload` callback out of inline definition into its own method for CSP compatibility [fb999e98e9cbaac67158689cc240d054c3318de4].

Enhancements:
- Alt tag: Adds `alt` tag into the images wherever available, from the `title` attribute in the API result [4ff8b908e2707e26f4eddec779d2a58e89c698dc].

### DIFF
--- a/plugins/giphy/trumbowyg.giphy.js
+++ b/plugins/giphy/trumbowyg.giphy.js
@@ -71,9 +71,11 @@
       .map(function (gifData) {
         var downsized = gifData.images.downsized || gifData.images.downsized_medium;
         var image = downsized,
-            imageRatio = image.height / image.width;
+            imageRatio = image.height / image.width,
+            altText = gifData.title;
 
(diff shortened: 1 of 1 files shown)


synth:
@Alex-D Indeed, this fixes 2 major bugs with the Giphy plugin as indicated in the description above. Would love to see this upstream so we can stop. 

NOTE: The first issue (image fallback) maps to issue  #1243. However, there is no existing issue for the CSP issue (happy to make one if necessary)

Alex-D (approved):

Alex-D:
Thank you :)

PR #1346, opened 2022-12-12
Fix modal focus on link/image add

35grain:
When the editor is displayed within a Bootstrap modal, for example, the additional Trumbowyg modal inputs are not focused properly. This fix appends the modal box right after the Trumbowyg box so that hierarchically it can receive focus without issues.

### DIFF
--- a/src/trumbowyg.js
+++ b/src/trumbowyg.js
@@ -1557,7 +1557,7 @@ Object.defineProperty(jQuery.trumbowyg, 'defaultOptions', {
             }).css({
                 top: t.$box.offset().top + t.$btnPane.height(),
                 zIndex: 99999
-            }).appendTo($(t.doc.body));
+            }).insertAfter($(t.$box));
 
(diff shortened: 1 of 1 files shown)


Alex-D:
I understand the fix, but I do not remember why it has been changed to the bottom of the body in the past :/

Alex-D:
Ok, after finding out, it seems like the issue and PRs were: 
- https://github.com/Alex-D/Trumbowyg/issues/775
- https://github.com/Alex-D/Trumbowyg/pull/779

So, your solution is not working either since it results again in a form inside another form :/

35grain:
I see. Not sure what the correct solution is then to be fair /: Perhaps adding an option for selecting the parent form element AFTER which the Input Links modal should be inserted, otherwise expect default behaviour? This way it stays within the parent modal and is valid HTML.

Alex-D:
It can be done in another commit/branch/PR since the right approach differs.

Issue #1119, opened 2020-05-14
SVG icon for custom plugin


mapreal19:
I'm working on a custom plugin for inserting a page break, but is not quite ready for sharing with others. Is it possible to provide a custom SVG when creating the button definition?

Currently I'm just showing a text instead of a button.

```js
var pageBreakContent = '<div class="no-print" style="page-break-before: always;">{{Page Break}}</div>';

var btnPageBreak = {
    title: 'Page Break',
    text: 'Page Break',
    hasIcon: false,
    fn: function () {
        trumbowyg.execCmd('insertHTML', pageBreakContent)
    }
};
```

Regarding the page break feature, when I save the HTML in the server I add a style tag with the style `<style>.no-print { opacity:0;}</style>`  before printing the page, but I guess we can have a better solution without relying on the server. Just a hack that works for me now. 

Related: https://github.com/Alex-D/Trumbowyg/issues/814

Alex-D:
Use `@media print { .no-print... }`

For the icon, check other plugins like giphy or upload, there is a folder with the SVG in it. It will embed at build time :)

mapreal19:
I tried to follow the structure of other plugins:

![image](https://user-images.githubusercontent.com/3055997/82435012-fa64c700-9a93-11ea-8c59-3c68f30fbf98.png)

And import the plugin as stated in the docs:

```
<script src="https://cdnjs.cloudflare.com/ajax/libs/Trumbowyg/2.20.0/trumbowyg.min.js"></script>
<script src="/static/plugin/pagebreak/trumbowyg.pagebreak.js"></script>
```

The plugin works, is just the icon doesn't load. Do I need to be more specific somewhere? Or use a bundler like webpack? I'm using the CDN for now just for demo purposes.

Alex-D:
You need to build Trumbowyg, all plugin icons are embed in the core :)

So you can't use CDN in this case.

mapreal19:
@Alex-D I was able to build the library with Webpack, but adding a custom icon wasn't that easy. Maybe there should be a guide on how to use it with Webpack?

lewiscollard:
Obviously, I am about three years late to the party here. But I'll leave my _less good_ solution to this for anyone in the future who is writing their own plugin.

[...]

PR #1362, opened 2023-01-20
Document adding custom button icons for standalone plugins.

lewiscollard:
The previously [recommended](https://github.com/Alex-D/Trumbowyg/issues/1119#issuecomment-631452561) way of adding a custom icon to a plugin's button was to roll one's own build of Trumbowyg, but for lazy people like me, that looks too much like effort.

Moreover, that implies that plugins _cannot_ be distributed standalone (e.g. load Trumbowyg from a CDN, then your own plugin on top of that). This is not so! But that notion might put someone off using Trumbowyg for their project. Trumbowyg doesn't actually care where the button's symbol comes from; if a symbol exists in the document with the correct ID it will work. So, this PR documents the "shove an SVG into the DOM" solution.

P.S. Thanks for Trumbowyg. It's awesome!

### DIFF
--- a/docs/documentation/plugins/index.html
+++ b/docs/documentation/plugins/index.html
@@ -1484,39 +1484,42 @@ fileToUpload: // The image file
 
     // If the plugin is a button
     function buildButtonIcon() {
-        if (!document.querySelector("#trumbowyg-myplugin")) {
-            /*
-            When your button is created, an SVG will be inserted with an xref
-            to a symbol living at the fragment "#trumbowyg-myplugin".
-            For Trumbowyg's own plugins, this will come from a sprite sheet
-            which is injected into the document, built from the icon SVG
-            in "ui/icons" in your plugin's source tree.
-
-            But, nothing says it *has* to come from Trumbowyg's injected
-            sprite sheet; it only requires that this symbol exists in your
-            document. To allow stand-alone distribution of your plugin, we're
-            going to insert a custom SVG symbol into the document with the
-            correct ID.
-            */
-            const iconWrap = $(document.createElementNS("http://www.w3.org/2000/svg", "svg"));
-
-            /*
-            This class is only here to allow CSS to hide it. In your plugin's
-            stylesheet you will probably want something like this:
-
-            .trumbowyg-myplugin-icons {
-                display: none;
-            }
-            */
-            iconWrap.addClass("trumbowyg-myplugin-icons");
-            // For demonstration purposes, we've taken the "File" icon from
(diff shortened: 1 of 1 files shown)


docs/documentation/plugins/index.html
```diff
@@ -1482,6 +1482,44 @@ <h3 id="skeleton">Skeleton</h3>
...
+            const iconWrap = $(document.createElementNS("http://www.w3.org/2000/svg", "svg"));
+
+            /*
+            This class is only here to allow CSS to hide it. In your plugin's
+            stylesheet you will probably want something like this:
+
+            .trumbowyg-myplugin-icons {
+                display: none;
+            }
+            */
```
Alex-D, line 1510: Maybe I should add a generic class like that to the trumbowyg CSS file, allowing it to be shared by the default SVG *and* all additional ones :)

docs/documentation/plugins/index.html
```diff
@@ -1482,6 +1482,44 @@ <h3 id="skeleton">Skeleton</h3>
         }
     }
 
+    // If the plugin is a button
+    function buildButtonIcon() {
+        if (!document.querySelector("#trumbowyg-myplugin")) {
```
Alex-D, line 1487: I prefer the early return way, aka if the element exists, return; then, the rest of the function runs only if we need to create the button. It prevent from having the whole function in a if

Alex-D (changes requested):

docs/documentation/plugins/index.html
```diff
@@ -1482,6 +1482,44 @@ <h3 id="skeleton">Skeleton</h3>
...
+            This class is only here to allow CSS to hide it. In your plugin's
+            stylesheet you will probably want something like this:
+
+            .trumbowyg-myplugin-icons {
+                display: none;
+            }
+            */
+            iconWrap.addClass("trumbowyg-myplugin-icons");
+            // For demonstration purposes, we've taken the "File" icon from
+            // Remix Icon - https://remixicon.com/
```
Alex-D, line 1513: Nice one, did not know it :)

docs/documentation/plugins/index.html
lewiscollard: Now that you mention it, so do I. See: fa7ead9

docs/documentation/plugins/index.html
lewiscollard: I think that is a good idea. Is it a blocker for this PR?

Alex-D (approved):

Alex-D:
Thank you for your suggestion :)

PR #1364, opened 2023-01-23
Add spelling fix change event handling

### LINKED ISSUE #1363: Fire change event when spelling suggestions are applied
### Description
This seems to be similar to #162 (if not the same issue) in that spelling changes (selected from the context menu) don't trigger any sort of event. However it does seem that the `input` event is fired on the editor element and when filtering this to events where the original event `data` is null it *seems* to provide a point where we can fire the `tbwchange` event, thus solving the problem. 

My first crack at this problem tried testing the `originalEvent.inputType` to check if it was `'insertReplacementText'` however Safari just uses the normal `'insertText'` value (Safari is the new IE, right?! 😆). However, the `data` property of the `originalEvent` does appear to consistently be `null` for the spelling change event.

ie
```js
.on('input', function (e) {
    // This worked fine on all browsers except Safari
    // if (e.originalEvent instanceof InputEvent && e.originalEvent.inputType === 'insertReplacementText') {

    // This worked on all tested browsers.
    // Although I couldn't say what other interactions with the input might trigger something similar??!
    if (e.originalEvent instanceof InputEvent && e.originalEvent.data === null) {
        t.$c.trigger('tbwchange');
    }
})
```

Tested against Chrome 108, Safari 16.2, Firefox nightly (v111), Edge 109, Opera 94

Happy to put together a pull request if the above sounds good 😄

abrom:
All modern browsers trigger the `input` event and assign `e.originalEvent.inputType` to `'insertReplacementText'` when applying spelling fixes, except for Safari which uses `e.originalEvent.inputType` of `'insertText'` and `e.originalEvent.data` of `null`.

With this is mind we can have the `twbchange` Trumbowyg event triggered on the editor when a spelling fix from the browser context menu is applied.

To test this, insert a textarea on a page and initialize, listening to the `twbchange` event. Without the fix applying a spelling fix will not see the `twbchange` event fired. With the fix you see see the `twbchange` event fired

fixes #1363

### DIFF
--- a/src/trumbowyg.js
+++ b/src/trumbowyg.js
@@ -662,6 +662,14 @@ Object.defineProperty(jQuery.trumbowyg, 'defaultOptions', {
                         ctrl = false;
                     }, 50);
                 })
+                .on('input', function (e) {
+                    // Trigger change event when spelling fixes applied
+                    var event = e.originalEvent;
+                    if (typeof event === 'object' && (event.inputType === 'insertReplacementText' ||
+                        (event.inputType === 'insertText' && event.data === null))) {
+                        t.$c.trigger('tbwchange');
+                    }
+                })
                 .on('mouseup keydown keyup', function (e) {
                     if ((!e.ctrlKey && !e.metaKey) || e.altKey) {
                         setTimeout(function () { // "hold on" to the ctrl key for 50ms


Alex-D (approved):

Alex-D:
Thanks :)

abrom:
No, worries! Thank you for such a great project 😄

PR #1394, opened 2023-04-09
 i18n: add / improve de translation by @johangroe

johangroe:
I improved and added some german translations for some plugins and the core.
Would be happy about a merge and a mention in the contributors file.


Alex-D (approved):

Alex-D:
Thank you :)

PR #1404, opened 2023-06-07
Fix IME input

hidemikimura:
Fixed a problem in which the style of span element disappears when IME input is confirmed.

### DIFF
--- a/src/trumbowyg.js
+++ b/src/trumbowyg.js
@@ -600,7 +600,7 @@ Object.defineProperty(jQuery.trumbowyg, 'defaultOptions', {
                 .on('keydown', function (e) {
                     // append flags to differentiate Chrome spans
                     var keyCode = e.which;
-                    if (keyCode === 8 || keyCode === 13 || keyCode === 46) {
+                    if (keyCode === 8 || keyCode === 13 || keyCode === 229 || keyCode === 46) {
                         t.toggleSpan(true);
                     }
                     if ((e.ctrlKey || e.metaKey) && !e.altKey) {


Alex-D:
Hi!
What does this fix exactly and how can I test the fix?

Alex-D:
I've read some documentation that seems related to the dead key.

Thank you for your contribution :)

KohtaHorike:
Hi @Alex-D 👋 — a heads-up that although this PR shows as **Merged**, the fix never actually shipped.

The merge commit `3c9fc78b` (2025-01-29 16:32 UTC) is no longer part of `develop`'s history: at 17:22 UTC the same day, `7d56742f` was pushed with parent `3ff8d569` — the *pre-merge* tip — so the merge commit and @hidemikimura's fix commit `09203990` were both dropped from the branch (presumably an accidental force-push from a local clone that predated the merge). You can verify with `git log develop` (neither commit is reachable) or by comparing `develop...3c9fc78b`.

As a result, v2.30.0 (released later that day), v2.31.0 and current `develop` all lack the fix, and the IME bug is still reproducible on the latest version: with a Japanese/Chinese/Korean IME, pressing Backspace during composition fires keydown with `keyCode === 229`, so existing `<span>` elements never get flagged by `toggleSpan(true)`, and the subsequent keyup unwraps every span in the editor — styled spans (e.g. pasted `<span style="font-weight: 700;">…</span>`) silently lose their formatting.

I've opened #1685, which cherry-picks the original commit onto current `develop` (preserving @hidemikimura's authorship). It would be great to get this landed again — thanks for maintaining Trumbowyg!

Issue #1385, opened 2023-03-11
After upgrading to the most recent version, my trumbowyg editor is now as long as the content


jeremytunnell:
Everything was solid and then I upgraded, and now the trumbowyg editor expands to the length of the content.  For some of my blog posts, the editor is 5000px tall.  

I have added:
autogrow: false,
autogrowOnEnter: false,

Neither of these helps.  Problem exists in both Brave and Firefox on Windows.

I have kind if hacky fixed the problem with:

<style>
.trumbowyg-editor {
max-width:700px;
max-height: 500px;
}
</style>

You can see an example of the problem here :

https://jeremytunnell.net/utils/test_tw

Can you recommend how to fix this correctly?

Thanks,
Jeremy

Alex-D:
I have seen this bug, but I have no room to fix it. For now, you can seed a max height. That is what was set before this update.

novotnicek:
Hello @jeremytunnell @Alex-D 

For "solid" height I using this CSS:

```css
.trumbowyg-box,
.trumbowyg-editor,
.trumbowyg-textarea {
    height: 400px;
}

.trumbowyg-box.trumbowyg-fullscreen,
.trumbowyg-box.trumbowyg-fullscreen .trumbowyg-editor,
.trumbowyg-box.trumbowyg-fullscreen .trumbowyg-textarea {
    height: 100%;
}
```

I don't know, if is really bug  :bug: :thinking:. Seems legit for me.

jeremytunnell:
Thanks @novotnicek.  That helps.  

It's still a bug though.  Default behavior should not be infinite height.

novotnicek:
Hello @jeremytunnell @Alex-D

I created PR #1418. @Alex-D pls, check it (it's short).

I added height options. Default is 'auto' and don't change current behaviour. Only when user set e.g. height: '300px', then trumbowyg don't "autogrow" with content.

novotnicek:
```js
$('.someClass textarea').trumbowyg({
    height: '300px',
});
```

sam-later:
New here.  I added a PR to fix this.  I think this got broken in v 2.28 and later.
https://github.com/Alex-D/Trumbowyg/pull/1643

PR #1418, opened 2023-08-23
#1385 add height option

novotnicek:
I added height options. Default is 'auto' and don't change current behaviour. Only when user set e.g. `height: '300px'`, then trumbowyg don't "autogrow" with content.

See #1385

### DIFF
--- a/src/trumbowyg.js
+++ b/src/trumbowyg.js
@@ -80,6 +80,7 @@ Object.defineProperty(jQuery.trumbowyg, 'defaultOptions', {
         autogrowOnEnter: false,
         imageWidthModalEdit: false,
         hideButtonTexts: null,
+        height: 'auto',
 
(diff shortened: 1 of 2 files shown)


MysticEntity-dev (commented):
had to remove line 573 and 575 in trumbowyg.js to get the wanted outcome.

src/trumbowyg.js
```diff
@@ -569,6 +570,10 @@ Object.defineProperty(jQuery.trumbowyg, 'defaultOptions', {
                 .html(html)
             ;
 
+            t.$box.css('height', t.o.height);
```
MysticEntity-dev, line 573:
```suggestion
            
```

Alex-D:
I don't think that's a good way to make the height go right. But I do not have time to check that for now :/

PR #1437, opened 2023-12-15
Add Giphy options for modularity with sensible, performant defaults

### LINKED ISSUE #1436: Giphy plugin causes slow image loading due to not using limit query parameter
Currently, the Giphy plugin does not include `limit` query parameter and can result in a really slow response from Giphy. I have seen this be upwards of 20s response time before images will load. 

In my testing, simply adding a limit parameter to something like 50 will result in sub 1s response times. 

This can be worked around by adding `limit` to the "rating" plugin option, but this is hacky. This plugin should support all the options that Giphy has to be future proof:

image

UPDATE: I originally thought just `limit` was the issue. In actuality it was more about the "renditions" used when the picker loads lots of gifs at once.

synth:
This adds more modularity to Giphy options while starting with sensible and performant defaults

closes #1436 

This uses the Giphy concept of "renditions" to add performant grids of giphs. 
Ref: https://developers.giphy.com/docs/optional-settings/#renditions-on-demand
Ref: https://developers.giphy.com/explorer/

### DIFF
--- a/plugins/giphy/trumbowyg.giphy.js
+++ b/plugins/giphy/trumbowyg.giphy.js
@@ -71,20 +71,18 @@
 
     var html = response.data
       .filter(function (gifData) {
-        // jshint camelcase:false
-        var downsized = gifData.images.downsized || gifData.images.downsized_medium;
-        // jshint camelcase:true
-        return !!downsized.url;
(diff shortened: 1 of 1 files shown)


Alex-D:
I've merged/rebased/improved your suggestions :)
Thank you for your contribution!

PR #1453, opened 2024-04-22
New Plugin: Speech recognition

Eruvaer74:
With this plugin Trumbowyg is able to use speech recognition.


Alex-D:
Thank you for your plugin!
That's a nice addition to the collection :)

I've tweaked it a bit to make it easier to use (avoid resetting the whole content each time we click on the button, etc.)

PR #1460, opened 2024-06-02
Update trumbowyg.history.js Tried to Fix the Undo and Redo Button which had issues

rockO7:
Fixed the Undo and Redo Button Functionality which had Issues when making it Bold and Then trying to Undo then entire Content was copying and Appending instead of removing the Bold One i have tried to Fix it please check and if any issues please let me kow


Alex-D:
From what I see in the code, this will only work if you change the text, but not if you change the style as you compare only the text instead of the tags in the old version.

The right way to do that should be something with both approaches.

PR #1468, opened 2024-06-24
xhr filelds is not being assigned in xhrRequest it was assigning headers.

yogi111:
### Expected
Xhr Fields should assigned to xhrFields in xhr request.

### Current Behaviour
Heades are being assigned to XhrFields

### Fix
replace the headers tp xhrFields


Alex-D (approved):

Alex-D:
Nice catch, thank you for the patch :)

PR #1493, opened 2024-10-24
Fixed table background and border color outside the editor

rdnasim:
I placed the `<textarea>` inside `<td>` and `<tr>` tags, but when I change the table's background and border color using the Trumbowyg toolbar, the changes are being applied outside the editor instead of to the table inside the editor.

I attempted to solve this issue with some simple changes, and it is now resolved. Please check the pull request to review the changes and get an idea of what was adjusted. 

I have also attached two images to provide a brief overview of the problem.

problem-background-color-table
problem-border-color-table

### DIFF
--- a/plugins/table/trumbowyg.table.js
+++ b/plugins/table/trumbowyg.table.js
@@ -1556,7 +1556,7 @@
                     };
                     var applyBackgroundColorToSelectedCells = function (color) {
                         return function () {
-                            var $table = $(t.doc.getSelection().anchorNode).closest('table');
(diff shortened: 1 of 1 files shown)


Alex-D:
Hey!
The simpler fix is to use the context in the closest function, as you can see here: 
https://github.com/Alex-D/Trumbowyg/commit/7d56742f63876f42dd63c2a36f9d8bdbd9fabf06

Thank you for you contribution that helps me to improve the plugin :)

PR #1657, opened 2026-02-23
feat (table): Add 'enableHeaderRow' option

Eren-Kaya:
I added the `enableHeaderRow` option to the table plugin which controls whether tables get built with header rows. The default value is set to true to match the current behavior. I added this option because in versions of Trumbowyg before v2.27.2 tables weren't created with header rows and some users may prefer the old behavior.

### DIFF
--- a/docs/documentation/plugins/index.html
+++ b/docs/documentation/plugins/index.html
@@ -1353,6 +1353,12 @@ td {
                     <br>
                     <strong>Default</strong>: <code>true</code>
                 </dd>
+                <dt><strong><code>enableHeaderRow</code></strong> <code class="type">boolean</code></dt>
+                <dd>
+                    Add a header row when building tables if enabled.
+                    <br>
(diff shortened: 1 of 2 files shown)


docs/documentation/plugins/index.html
```diff
@@ -1353,6 +1353,12 @@ <h4 id="plugin-table--options">Options</h4>
                     <br>
                     <strong>Default</strong>: <code>true</code>
                 </dd>
+                <dt><strong><code>enableHeaderRow</code></strong> <code class="type">boolean</code></dt>
```
Alex-D:
```suggestion
                <dt><strong><code>addHeaderRowToNewTables</code></strong> <code class="type">boolean</code></dt>
```

Maybe this name is more explicit?

Alex-D (changes requested):

Alex-D (approved):

Alex-D:
Thank you for this addition to the table plugin :)

Issue #1477, opened 2024-07-18
Add the ability to restrict image types


ataylor32:
The "base64" and "upload" plugins both have the `accept` attribute set to `image/*`:

https://github.com/Alex-D/Trumbowyg/blob/v2.28.0/plugins/base64/trumbowyg.base64.js#L150

https://github.com/Alex-D/Trumbowyg/blob/v2.28.0/plugins/upload/trumbowyg.upload.js#L212

It would be nice if this was configurable. Currently, any image type is allowed, including `.bmp`, `.ico`, and others. I would like to have the ability to restrict this to only certain image types.

PR #1670, opened 2026-04-10
feat(base64): add `mimeTypes` option

mpryvkin:
add #1477

### DIFF
--- a/docs/demos/plugins/base64.html
+++ b/docs/demos/plugins/base64.html
@@ -119,7 +119,7 @@ $('#editor-max-file-size')
 
             <div id="editor-mime-types">
                 <p>
-                    Option <code>mimeTypes</code> allows to specify allowed MIME types. Default value is <code>image/*</code> which accepts all images.
+                    Option <code>mimeTypes</code> allows to specify allowed MIME types. Default value is <code>['image/*']</code> which accepts all images.
                 </p>
             </div>
 
--- a/docs/demos/plugins/base64.html
+++ b/docs/demos/plugins/base64.html
@@ -115,6 +115,31 @@ $('#editor-max-file-size')
         </div>
 
         <div class="feature">
+            <h3>MIME types</h3>
+
+            <div id="editor-mime-types">
+                <p>
+                    Option <code>mimeTypes</code> allows to specify allowed MIME types. Default value is <code>image/*</code> which accepts all images.
+                </p>
+            </div>
+
+            <h4>The code</h4>
+            <pre><code class="js-code-to-eval javascript">
+$('#editor-mime-types')
+.trumbowyg({
+    btns: [
+        ['base64']
+    ],
+    plugins: {
+        base64: {
+            mimeTypes: ['image/gif', 'image/jpeg', 'image/png', 'image/webp']
+        }
+    }
+});
+            </code></pre>
+        </div>
+
+        <div class="feature">
             <h3>Setup</h3>
 
             <h4>In head tag</h4>
--- a/plugins/base64/trumbowyg.base64.js
+++ b/plugins/base64/trumbowyg.base64.js
@@ -187,7 +187,7 @@
                             trumbowyg.saveRange();
 
                             var file;
-                            var mimeTypes = trumbowyg.o.plugins.base64.mimeTypes || ['image/*'];
+                            var mimeTypes = trumbowyg.o.plugins.base64.mimeTypes;
 
                             var $modal = trumbowyg.openModalInsert(
                                 // Title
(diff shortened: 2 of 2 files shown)


plugins/base64/trumbowyg.base64.js
```diff
@@ -131,12 +151,16 @@
...
                 init: function (trumbowyg) {
+                    trumbowyg.o.plugins.base64 = trumbowyg.o.plugins.base64 || defaultOptions;
+
                     var btnDef = {
                         isSupported: isSupported,
                         fn: function () {
                             trumbowyg.saveRange();
 
                             var file;
+                            var mimeTypes = trumbowyg.o.plugins.base64.mimeTypes || ['image/*'];
```
Alex-D: Why is that not in the `defaultOptions` only?

plugins/base64/trumbowyg.base64.js
```diff
@@ -131,12 +151,16 @@
             base64: {
                 shouldInit: isSupported,
                 init: function (trumbowyg) {
+                    trumbowyg.o.plugins.base64 = trumbowyg.o.plugins.base64 || defaultOptions;
```
Alex-D: This should be done using the extends:


```suggestion
                    trumbowyg.o.plugins.base64 = $.extend({},
                        defaultOptions,
                        trumbowyg.o.plugins.base64 || {}
                    );
```

This way, you can get a deep copy.

plugins/base64/trumbowyg.base64.js
Alex-D:
```suggestion
                            var allowedMimeTypes = trumbowyg.o.plugins.base64.allowedMimeTypes;
```

plugins/base64/trumbowyg.base64.js
```diff
@@ -158,6 +182,25 @@
 
                                 // Callback
                                 function (values) {
+                                    // Validate MIME type
+                                    if (file && mimeTypes.indexOf('image/*') === -1) {
```
Alex-D: To be honest, I don't understand this part of the code.

Plus: the `errInvalidImage` i18n key is never used.

plugins/base64/trumbowyg.base64.js
```diff
@@ -158,6 +182,25 @@
...
+                                            if (mimeType.indexOf('/*') !== -1) {
+                                                return fileType.indexOf(mimeType.replace('/*', '/')) === 0;
+                                            }
+                                            return fileType === mimeType;
+                                        });
+
+                                        if (!isAllowed) {
+                                            trumbowyg.addErrorOnModalField(
+                                                $('input[type=file]', $modal),
+                                                trumbowyg.lang.errInvalidMimeType
```
Alex-D: Is this really invalid? Should be more something like _Not Allowed_ than _Invalid_.

Alex-D (changes requested):

plugins/base64/trumbowyg.base64.js
mpryvkin: This line checks if the file has been provided and the list of allowed MIME types doesn't have an entry that accepts all images (such as `image/*`, which is a default value). 
1. If all images are allowed (default value), file type will not be compared against provided MIME types.
2. If only specific MIME types are allowed , file type will be compared against provided MIME types.

Key `errInvalidImage` is used [here](
https://github.com/mpryvkin/Trumbowyg/blob/69c03b2888115641953bbb3610f9565fb3252f1c/plugins/base64/trumbowyg.base64.js#L215).

plugins/base64/trumbowyg.base64.js
mpryvkin: Agree, see commit 54589c89a8aacd026cbe3d249cf95dc15a1b934e

plugins/base64/trumbowyg.base64.js
mpryvkin: Agree, see commit ccc61fb1d31feefbd715fb1f0fc7d4e2da39a0cb

plugins/base64/trumbowyg.base64.js
mpryvkin: Agree. Improved handling of `mimeTypes` option, fixed typo with default option name, added explanation. See 6fc33105cdc6d8aaa3e4ea4bb840f80e897a8a24

plugins/base64/trumbowyg.base64.js
mpryvkin: I improved and documented the condition, see 1ecfd77657111439f3072826825e136834a0001c

docs/demos/plugins/base64.html
```diff
@@ -89,6 +89,31 @@ <h4>The code</h4>
             </code></pre>
         </div>
 
+        <div class="feature">
+            <h3>MIME types</h3>
+
+            <div id="editor-mime-types">
+                <p>
+                    Option <code>mimeTypes</code> allows to specify allowed MIME types. Default value is <code>image/*</code> which accepts all images.
```
Alex-D:
```suggestion
                    Option <code>mimeTypes</code> allows to specify allowed MIME types. Default value is <code>['image/*']</code> which accepts all images.
```

Alex-D (changes requested):
Aside that, it seems legit :)

Alex-D:
Since I've merged another PR, you will need to fix conflicts :/

mpryvkin:
@Alex-D, sorry for the delay, got sidetracked. Just rebased the commits to fix the conflicts. I noticed you prefer not to squash commits when merging, is that on purpose?

PR #1683, opened 2026-06-16
Preformatted plugin improvement

stefl0n:
The preformat-function is imho ment to wrap and **convert** a **pre-formatted text** (e.g. code example) to a proper `pre` or `code` block.

Current behaviour (on the trumbowyg homepage):
preformatted-1

Expected behavior (in this PR):
preformatted-2


plugins/preformatted/trumbowyg.preformatted.js
```diff
@@ -121,25 +127,19 @@
...
-    /*
-     * UnwrapCode
-     * ADD/FIX: to improve, works but can be better
-     * "paranoic" solution
-     */
     function unwrapCode() {
         var container = null;
 
-        if (document.selection) { //for IE
+        if (document.selection) {
```
Alex-D: Why removing the comment?

plugins/preformatted/trumbowyg.preformatted.js
```diff
@@ -121,25 +127,19 @@
...
-        var tmp = document.createElement('DIV');
-        tmp.innerHTML = html;
-        return tmp.textContent || tmp.innerText || '';
+    function convertToPreformatted(html) {
+        return html .replace(/<p>/gi, '')
+            .replace(/<\/p>/gi, '\n')
+            .replace(/<br\s*\/?>/gi, '\n')
+            .replace(/<h[1-6]>|<\/h[1-6]>/gi, '\n')
+            .replace(/<li>/gi, '- ')
+            .replace(/<\/li>/gi, '\n');
```
Alex-D: I'm unsure about this way to do it, there are a lot more tags missing here, no?
What's the goal of this function?

plugins/preformatted/trumbowyg.preformatted.js
```diff
@@ -148,16 +148,22 @@
...
+        // step 1: make html safe
+        let html = text
+            .replace(/&/g, '&amp;')
+            .replace(/</g, '&lt;')
+            .replace(/>/g, '&gt;');
+
+        // step 2: keep whitespaces & line breaks
+        html = html
+            .replace(/  /g, '&nbsp;&nbsp;')             // keep double spaces
+            .replace(/\t/g, '&nbsp;&nbsp;&nbsp;&nbsp;') // keep tabs
```
Alex-D: Why replacing tabs with 4 spaces? Can't we use `&Tab;`?

Alex-D (changes requested):
I've added some feedbacks, also, can you rebase on the main branch since I've merge your other PR?
Thank you :)

plugins/preformatted/trumbowyg.preformatted.js
stefl0n: OK, the comment may stay. I accidentally removed it while cleaning up my own comments.

plugins/preformatted/trumbowyg.preformatted.js
stefl0n: You are right. That was indeed not my final version of that conversion function. I rewrote it now completely in my final version.

plugins/preformatted/trumbowyg.preformatted.js
stefl0n: Yes. good point. Let's use &tab; here.

stefl0n:
I rebased with main and tried to commit my final changes. Not sure if everything went well 🙈

PR #1690, opened 2026-09-16
Preserve link style on edit

TFleury:
Link edition removes all existing style or CSS classes.
This PR just keeps existing styles and CSS classes and restores them after edit.


Alex-D (approved):

Alex-D:
Seems legit in fact :)

